PR #3, opened 2015-09-08
Frontend

ramirezg:
Adding 18f pages for documentation front end.


ramirezg:
screen shot 2015-09-08 at 12 27 23 pm

mogul:
The control listing in the table of contents at left should probably include the summary, eg "AC-2" should have "Account Management" next to it... People (like me!) won't have enough context from just the control code to make the TOC useful otherwise.

PR #7, opened 2015-09-22
Component-level yamls

ramirezg:
Started to transform an [intermediary spreadsheet](https://github.com/18F/control-masonry/blob/matrix-2/Control-masonry.xlsx) to help create the first batch of component-level control documentation.


README.md
```diff
@@ -1,90 +1,72 @@
...
+satisfies:
+  AC-2: Cloud Foundry accounts are managed through the User Account and Authentication
+    (UAA) Server.
+  IA-2: The UAA is the identity management service for Cloud Foundry. Its primary
+    role is as an OAuth2 provider, issuing tokens for client applications to use when
+    they act on behalf of Cloud Foundry users.
+  SC-13: All traffic from the public internet to the Cloud Controller and UAA happens
+    over HTTPS and operators configure encryption of the identity store in the UAA
+  SC-28 (1): Operators configure encryption of the identity store in the UAA. When
+    users register an account with the Cloud Foundry platform, the UAA, acts as the
```
joshuamckenty: I really think we need a hierarchy of "components" - UAA is a component of CF, but most government CF installations actually use the UAA plugin mechanism to connect external IDP systems such as CA Siteminder (because they fulfill all the 800-53 AC controls that way). So I think the "component" that satisfies AC-2 is a sub-component of UAA.

ramirezg: This is interesting. I guess I'm wondering how far down the chain we have to go. Do you think that all systems have sub-components that satisfy the controls level?  
Like this:
Cloud Foundry > UAA > CA Siteminder (AC-2)
AWS > EC2 > Something else (AC-3)
Or will there be cases where the component itself satisfies one control and the sub-component satisfies another?
Cloud Foundry > UAA (satisfies AC-3, IA-9, etc.) > CA Siteminder (satisfies AC-2)

PR #9, opened 2015-09-28
Improve content of NIST-800-53.yaml

gregelin:
I've update NIST-800-53.yaml with full text of control from NIST SP 800-53. The YAML file is generated 
[GovReady/800-53-server](https://github.com/GovReady/800-53-server) via the call
```
http://localhost:8080/controllist/?ids=AC-2,AC-3,AC-6,AU-2,AU-6,CA-8,CM-2,CM-3,CM-6,CM-7,IA-2,PL-8,RA-5,SA-22,SC-7,SC-13,SI-2,SI-4,SI-10&format=control-masonry
```

The content is parsed from https://github.com/GovReady/800-53-server/blob/master/data/800-53-controls.xml document

### DIFF
--- a/data/standards/NIST-800-53.yaml
+++ b/data/standards/NIST-800-53.yaml
@@ -1,11 +1,130 @@
-name: NIST-800-53
 AC-2:
-  name: Account Management
-  description: There is an affordance for managing access by...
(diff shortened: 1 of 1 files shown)


ramirezg:
:+1: Thanks for the data!

PR #12, opened 2015-10-01
Processing NIST-800-53.yaml

ramirezg:
- Removed `&colon;` and `\t` from `NIST-800-53.yaml`
- Updated the FISMA documentation by running `yamls_to_certification.py`
- Small PEP8 related changes to `yamls_to_certification.py`

TODO incorporate more of  `NIST-800-53.yaml` into final documentation.


gregelin:
@ramirezg 

Can you say a bit more about removing the `&colon`? I wasn't sure the best way to address `:` in the text of the 800-53 control.

ramirezg:
@gregelin good point. I just removed it to make it more readable for people looking at the yamls, but I'll add back because it does render nicely on markdown.

PR #20, opened 2015-10-13
Closes #17. Check for info keys for listed controls.

### LINKED ISSUE #17: certifications_to_pages.py crashes from KeyError on controls listed in certifications/FISMA.yaml but not in standards/NIST-800-53.yaml
Running `python renderers/certifications_to_pages.py` crashes from KeyError when control is listed in `data/certifications/FISMA.yaml` but control not listed in `standards/NIST-800-53.yaml`. 

Since it is likely that people will have some trouble keeping everything in sync, it would be better if exceptions are caught and logged to standout like `yaml_to_certifications.py` is already doing.

Recreating error is easy. Add a control to `data/certifications/FISMA.yaml` like "AU-19" that does not exist in  `standards/NIST-800-53.yaml` and then run `python renderers/yaml_to_certifications.py` followed by running `python renderers/certifications_to_pages.py`

```
Gregs-MacBook-Pro:dsca-control-masonry greg$ python renderers/certifications_to_pages.py
Traceback (most recent call last):
  File "renderers/certifications_to_pages.py", line 144, in <module>
    output_path="exports/Pages"
  File "renderers/certifications_to_pages.py", line 133, in convert_certifications
    create_control_nav(control_key, control)
  File "renderers/certifications_to_pages.py", line 44, in create_control_nav
    'text': control_key + " " + control['meta']['name'],
KeyError: 'meta'
Gregs-MacBook-Pro:dsca-control-masonry greg$ python renderers/certifications_to_pages.py
Traceback (most recent call last):
  File "renderers/certifications_to_pages.py", line 144, in <module>
    output_path="exports/Pages"
  File "renderers/certifications_to_pages.py", line 133, in convert_certifications
    create_control_nav(control_key, control)
  File "renderers/certifications_to_pages.py", line 44, in create_control_nav
    'text': control_key + " " + control['meta']['name'],
KeyError: 'meta'
```

gregelin:
Modified `renderers/certifications_to_pages.py` to check if `meta` key existed for a control before trying to look up control name. Replaces missing name with zero length string and generates log error to standout out so pages will still be generated.

Unfortunately, this solution does result in missing pages for a control with no name.

### DIFF
--- a/data/certifications/FISMA.yaml
+++ b/data/certifications/FISMA.yaml
@@ -4,5 +4,6 @@ standards:
     AC-2:
     AC-6:
     CM-2:
+    AU-19:
   PCI:
     Regulation-6:
--- a/exports/Pages/_config.yml
+++ b/exports/Pages/_config.yml
@@ -32,6 +32,9 @@ navigation:
   url: PCI/
 - children:
   - internal: true
+    text: 'AU-19 '
+    url: AU-19/
+  - internal: true
     text: AC-2 ACCOUNT MANAGEMENT
     url: AC-2/
   - internal: true
--- a/exports/certifications/FISMA.yaml
+++ b/exports/certifications/FISMA.yaml
@@ -83,6 +83,7 @@ standards:
           of users) which are necessary to accomplish assigned tasks in accordance
           with organizational missions and business functions.
         name: LEAST PRIVILEGE
+    AU-19: {}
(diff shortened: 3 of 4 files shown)


ramirezg:
:+1: looking over this pr made me realize that I should actually split `create_markdown` into two different function so all the errors show.

PR #34, opened 2015-10-19
Add VPC component data

jcscottiii:
Removed CM-2 section as it is more Cloud Formation info
Added details to SC-7b for VPC


jcscottiii:
Wrong branch to merge into, refer to #35

PR #44, opened 2015-10-19
Update visualops component

jcscottiii:
No new information, just update layout


jcscottiii:
Refer to #44 for correct branch

PR #87, opened 2015-11-17
Adding the PCI-DSS Standard in the same yaml format

Gergues:
Added the PCI-DSS-MAY2015 standards in the same yaml format used for the NIST-800-53 standard.
Adapted from this source.
https://www.pcisecuritystandards.org/documents/Prioritized_Approach_v3-1.xlsx

### DIFF
--- a/src/templates/standards/PCI-DSS-MAY-2015.yaml
+++ b/src/templates/standards/PCI-DSS-MAY-2015.yaml
@@ -1,0 +1,523 @@
+name: PCI-DSS-MAY-2015
+Requirement 1:
+  name: Install and maintain a firewall configuration to protect cardholder data
+1.1:
+  name: Establish and implement firewall and router configuration standards.
+1.1.1:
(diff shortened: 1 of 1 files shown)


joshuamckenty:
Sweet!

mogul:
Wow, thanks a ton!! Now I think if someone from HHS contributes HIPAA we'll be in great shape...

geramirez:
This is fantastic, @Gergues. Bear with us as we update the schema. 

Also would you mind sharing how you converted from xlsx to yaml?

Gergues:
The conversion was done via Excel, or office libre  calculations.   very simple text replacement and concatenation.

Validation via www.yamllint.com .

Do you want that excel sheet calculations???  
Should i keep them in the same folder ,or under some other folder ??? 

-----Original Message-----

geramirez:
@Gergues we just wanted to see if there was a repeatable way of getting the data for when PCI is updated. In any case! Thanks for the contribution! Looks like it's working with a few tweaks to our script on our end. Thanks!

PR #100, opened 2016-02-05
In readme, put example on top and link to benefits

brittag:
To help new readers who are unfamiliar with the Compliance Masonry project, here's a suggestion to put the cloud.gov example at the top for easy clicking, along with a quick link to the Benefits explanation. I also added a little more detail to the example and made two tiny copyedits.


geramirez:
Thanks! @brittag

PR #105, opened 2016-02-17
Updating Compliance Masonry to work with v2 OpenControl schema

geramirez:
- Adding a transformation from v1 to v2
- Validating v2 files. 
- moving to version 2.0 to keep in pace with the OpenControl schema.


mzia:
LGTM

PR #108, opened 2016-04-11
Migrate from compliance-masonry-go to compliance-masonry

jcscottiii:
This PR:
- Removes all the python stuff from the repo
- Puts all the stuff from compliance-masonry-go into this repo.
- Maintains history from both repos.
@geramirez @ctro


geramirez:
Looks like you also changed the import paths. :+1: Looks good to me, we should also update this with the most recent merges.

PR #117, opened 2016-04-12
Possible docx template example [Do not merge]

geramirez:
Currently the template engine only supports one way of getting control data. 
For example the tag below produces
```txt
{{ getControl "NIST-800-53@CM-2"}}
```
this:
```txt
Amazon Elastic Compute Cloud  
Justification in narrative form  
Covered By:  
- EC2 Verification 1 http://VerificationURL.com  
```

We could implement more complex functions in templates using code similar to the code in this branch to customize the way data is rendered. 
For example:
```txt
{{ with getJustifications “NIST-800-53@CM-2” }}
{{ range $justification := . }}
{{$justification}}
{{$justification.SatisfiesData.Narrative}}
{{len $justification.SatisfiesData.CoveredBy }}
{{range $coveredBy := $justification.SatisfiesData.CoveredBy}}
{{$coveredBy}}
{{end}}
{{end}}
{{end}}

```

### DIFF
--- a/docx/docx.go
+++ b/docx/docx.go
@@ -5,8 +5,8 @@ import (
 	"strings"
 	"text/template"
 
-	"github.com/opencontrol/doc-template"
 	"github.com/opencontrol/compliance-masonry/models"
+	"github.com/opencontrol/doc-template"
 )
 
 // Config contains data for docx template export configurations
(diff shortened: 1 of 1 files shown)


afeld:
@jcscottiii Should we keep this PR open, or has it been subsumed by other changes?

PR #119, opened 2016-04-12
fix url to github.com/18F/cg-compliance

ajb:


afeld:
Not sure why the build is failing, but the change looks good to me. @jcscottiii? Thanks Adam!

PR #120, opened 2016-04-13
Move to using codecov

jcscottiii:
For fork PRs, circleci will suppress the token to upload a coverage report. Codecov allows you to upload the report without a token because it has internal apis circleci.

Also, coveralls does not seem to be working recently as smoothly as usual lately.


jcscottiii:
@afeld @ctro @geramirez

PR #121, opened 2016-04-15
Link to new blog post in readme

brittag:
The blog post has great explanation and background info! Also spelling out CLI, since government and tech have so many abbreviations and acronyms to keep track of.


afeld:
Thanks!

PR #127, opened 2016-04-27
Specify goxc tasks and exclude ones that take don't exclude vendor

jcscottiii:


.goxc.json
```diff
@@ -1,10 +1,23 @@
...
-		"default",
-		"publish-github"
-	],
 	"BuildConstraints": "linux,windows,darwin",
 	"PackageVersion": "1.0.1",
+	"MainDirsExclude": "./vendor",
+	"TasksExclude": [
+		"go-test",
+		"go-vet"
+	],
```
afeld: We wouldn't want to include these?

afeld:
Is there a way to test this new config locally without publishing a new release?

afeld:
Talked to @jcscottiii in person, and it sounds like we're going to try a different approach.

/cc https://github.com/laher/goxc/issues/99

Issue #133, opened 2016-05-09
As a user, I should be able to use CM to get a current gap analysis


DavidEBest:
* [ ] CM can specify and consume RequiredControls
* [ ] CM can take Inventory of its current Controls
* [ ] CM can compare RequiredControls and Inventory and output GapAnalysis

afeld:
> CM can specify and consume RequiredControls

Mind explaining this? @jcscottiii @ctro and I saw that added to the schema in #112 but weren't sure why it actually needs to be present in the YAML. What constitutes a non-required control?

ctro:
I think that these requirements came from Gabe, specifically "specify and consume Required Controls". 

This is why the one commit I made to gap analysis added that YAML.  Gabe envisioned adding all "required" controls to the project's yaml.  Then, you can check for the existence of those requiredControls in your actual documentation.  

Is this because projects don't want/need to implement ALL of the controls for some Certification??  In this case diffing against all controls wouldn't be too useful.  

I am also unsure of what makes a control required or not.

Maybe defining your list of requiredControls is the first step of using CM.  How many controls are usually required for a project though?  10? 30? 400?

afeld:
All good questions. Maybe this is related to some controls being partially fulfilled by the different layers, e.g. that the platform itself has monitoring, but that is not the same thing as an application having monitoring. @mogul @dlapiduz Any guesses?

mogul:
The required controls are determined by the baseline you're shooting for. So if you're trying to document a FISMA Moderate system, that will imply a certain set of controls need to be satisfied, which means that you should have something showing up under each of them in your assembled docs. The gap report should say "To get FISMA Moderate, there should be something here, here, and here". If it were FISMA Low, it'd be a smaller set. If it was PCI or HIPAA, it would be a different set. Make sense?

afeld:
Right...but doesn't a control being listed in the certification make it implicitly "required"?

https://github.com/opencontrol/schemas#certifications

PR #141, opened 2016-05-12
Gap analysis

jcscottiii:
Closes https://github.com/opencontrol/compliance-masonry/issues/133.

* Add section to README for Gap Analysis
* Move getCertifications to tools package as a helper function
* Add unit Tests
* Add entrypoint to the command line called `diff` to perform analysis

cc: @afeld @mogul 

image


README.md
```diff
@@ -111,6 +111,18 @@ The `gitbook` command by default will create a folder called `exports` that cont
...
+
+1. Once you collect the dependencies via the `get` command, run:
+
+    ```bash
+    $ compliance-masonry diff
+    Number of missing controls: 3
+    AT-4
+    AU-12
+    ...
+    ```
```
afeld: We should expand this section:

* What is gap analysis?
* When would you use this command? (at the start, and continually as you work)
* What are the things it tells you? (include more complete output)

inventory/inventory.go
```diff
@@ -0,0 +1,75 @@
...
+		for _, actualControl := range i.actualSatifiedControls {
+			if masterControl.EqualToSatisfiedControl(actualControl) {
+				found = true
+				break
+			}
+		}
+		if !found {
+			i.MissingControlList = append(i.MissingControlList, masterControl)
+		}
+	}
```
afeld: This could be a lot simpler by using a Set package, e.g. (first thing that came up in Google) https://github.com/deckarep/golang-set.

README.md
jcscottiii: I agree. There was little problem that I came across and how I'm going to still fix this inefficiency in the next commit.

I'm anticipating that in the future we will want to do more detailed gap analysis (e.g. show gap analysis for which controls are marked as "partially" or "not" implemented. [example](https://github.com/18F/cg-compliance/blob/a8c118a91ce61017c913948d4ebc063a0a556158/AC_Policy/component.yaml#L13)) In order to do that, I'm going to need the full object and not just unique strings. Which is all good but the type of object in `masterControlList` is of type `models.Control` and the ones given to us by the ones actually documented, `actualSatisfiedControls`, are of `models.Satisfies`. Which won't work out the box with set algebra.

New solution
```go
type Inventory struct {
	*models.OpenControl
	masterControlList       map[string]models.Control
	actualSatisfiedControls map[string]models.Satisfies
	MissingControlList      map[string]models.Control
}
```
The keys for `masterControlList` and `actualSatisfiedControls` are those unique strings with the `standard@control` but each map will contains their respective object as the value. I can do quick look ups and comparisons without the double for-loop and in the future we can pull out more details from each of the objects.

inventory/inventory.go
jcscottiii: I agree. There was little problem that I came across and how I'm going to still fix this inefficiency in the next commit.

I'm anticipating that in the future we will want to do more detailed gap analysis (e.g. show gap analysis for which controls are marked as "partially" or "not" implemented. [example](https://github.com/18F/cg-compliance/blob/a8c118a91ce61017c913948d4ebc063a0a556158/AC_Policy/component.yaml#L13)) In order to do that, I'm going to need the full object and not just unique strings. Which is all good but the type of object in `masterControlList` is of type `models.Control` and the ones given to us by the ones actually documented, `actualSatisfiedControls`, are of `models.Satisfies`. Which won't work out the box with set algebra.

New solution
```go
type Inventory struct {
	*models.OpenControl
	masterControlList       map[string]models.Control
	actualSatisfiedControls map[string]models.Satisfies
	MissingControlList      map[string]models.Control
}
```
The keys for `masterControlList` and `actualSatisfiedControls` are those unique strings with the `standard@control` but each map will contains their respective object as the value. I can do quick look ups and comparisons without the double for-loop and in the future we can pull out more details from each of the objects.

README.md
```diff
@@ -111,6 +111,23 @@ The `gitbook` command by default will create a folder called `exports` that cont
 
 Running the `docx` command will by default create a file named `export.docx` in the local directory.
 
+## Gap Analysis
+
+Use Gap Analysis to determine the difference between how many controls you have documented versus the total controls for a given certification. This should be used continually as you work to indicate your compliance progress.
+
+1. Given you have an `opencontrol.yaml` for your project and you have already collected your dependencies via the `compliance-masonry get` command, run `compliance-masonry diff <the-certification>`:
```
afeld: Minor: no need for the `1.` here.

inventory/inventory.go
```diff
@@ -0,0 +1,72 @@
...
+type Inventory struct {
+	*models.OpenControl
+	masterControlList       map[string]models.Control
+	actualSatisfiedControls map[string]models.Satisfies
+	MissingControlList      map[string]models.Control
+}
+
+// standardAndControlString makes a string from the standard and the control.
+// This is helpful for functions that want to create unique keys consistently.
+func standardAndControlString(standard string, control string) string {
```
afeld: Should this accept a `models.Control` instead?

inventory/inventory.go
```diff
@@ -0,0 +1,72 @@
...
+			}
+		}
+	}
+
+	// Missing controls
+	for standardAndControlKey, control := range i.masterControlList {
+		if _, exists := i.actualSatisfiedControls[standardAndControlKey]; !exists {
+			i.MissingControlList[standardAndControlKey] = control
+		}
+	}
```
afeld: Maybe move each of these loops into a method?

inventory/inventory_test.go
```diff
@@ -0,0 +1,61 @@
...
+			Context("When bad / no folder location is given", func() {
+				It("should return an empty slice and an error", func() {
+					config := Config{Certification: "LATO"}
+					i, err := ComputeGapAnalysis(config)
+					assert.Equal(GinkgoT(), []string{"Error: `certifications` directory does exist"}, err)
+					assert.Equal(GinkgoT(), 0, len(i.MissingControlList))
+				})
+			})
+		})
+		Context("When there controls specified in the certification but no controls have been documented", func() {
```
afeld: Should be "when there _are_ controls".

masonry-go.go
```diff
@@ -158,6 +159,35 @@ func NewCLIApp() *cli.App {
...
+				inventory, err := inventory.ComputeGapAnalysis(config)
+				if err != nil && len(err) > 0 {
+					fmt.Println(strings.Join(err, "\n"))
+					os.Exit(1)
+				}
+
+				fmt.Printf("\nNumber of missing controls: %d\n", len(inventory.MissingControlList))
+				for standardAndControl, _ := range inventory.MissingControlList {
+					fmt.Println(standardAndControl)
+				}
```
afeld: Since it doesn't seem easy to test this file, maybe move all of this into a standalone function where we can test the output?

inventory/inventory.go
```diff
@@ -0,0 +1,72 @@
...
+	// Master Controls
+	for standardKey, standard := range i.Certification.Standards {
+		for controlKey, control := range standard.Controls {
+			key := standardAndControlString(standardKey, controlKey)
+			if _, exists := i.masterControlList[key]; !exists {
+				i.masterControlList[key] = control
+			}
+		}
+	}
+	// Actual Controls
```
afeld: Maybe we call these "documented controls"?

afeld:
Looks great! Just a few relatively small things.

README.md
jcscottiii: Will do

inventory/inventory.go
jcscottiii: both model.Control and model.Satisfies pass in their respective standard and and control to this function.
A couple of alternatives: 
- We need to create two `standardAndControlString` functions `standardAndControlString(model.Control)` and `standardAndControlString(model.Satisfies)`.
- But I think at that point we would make those functions part of the struct instead. e.g. `func (c Control) String() string` and `func (s Satisfies) String() string`
- Or leave as is.

I think I will do the second one. For long term maintainability and style

inventory/inventory_test.go
jcscottiii: Good catch!

masonry-go.go
jcscottiii: Great idea!

inventory/inventory.go
jcscottiii: change of plans. Going with option 3. when creating the standard-control key with `models.Satisfies` for the actual documented controls, I just need that one object. With the master list, not only do i need `models.Control`, but i also need the enclosing struct. And that enclosing struct contains many Controls. So it gets a little complicated.

afeld: Okey doke. We can take a pass at refactoring later.

inventory/inventory.go
jcscottiii: will do!

jcscottiii: done

jcscottiii: it shouldve outdated this diff

masonry-go.go
jcscottiii: it shouldve outdated this diff

jcscottiii:
@afeld done! take a look

diff_test.go
```diff
@@ -0,0 +1,42 @@
...
+				assert.Contains(GinkgoT(), errBuffer.String(), "Error: Missing Certification Argument")
+			}
+		})
+		It("should return return the number a missing controls when given a certification", func() {
+			err := app.Run([]string{app.Name, "diff", "LATO", "-o", filepath.Join("fixtures", "opencontrol_fixtures")})
+			assert.Nil(GinkgoT(), err)
+			assert.Contains(GinkgoT(), buffer.String(), "Number of missing controls:")
+		})
+	})
+})
```
afeld: Niiice.

Issue #136, opened 2016-05-09
Add 'Not Implemented' boilerplate to incomplete controls in GitBook output.


DavidEBest:
Right now, controls that are not implemented are just blank, like:
https://compliance.cloud.gov/standards/NIST-800-53-CM-4.html

They should have some (maybe configurable) text, like 'Not Implemented', or 'This space left intentionally blank.'

PR #143, opened 2016-05-17
Narrative section

jcscottiii:
This addresses https://github.com/18F/compliance-toolkit/issues/25

## Updated templates and output 5/27

Template example:
image

Final example:
image


fixtures/component_fixtures/EC2/component.yaml
```diff
@@ -12,7 +12,11 @@ satisfies:
     system_key: CloudFoundry
     verification_key: UAA_Verification_1
   implementation_status: partial
-  narrative: Justification in narrative form
+  narrative:
+    - key: "a"
+      text: "Justification in narrative form A for CM-2"
+    - key: "b"
+      text: "Justification in narrative form B for CM-2"
```
afeld: Do we need to have a corresponding [schema](https://github.com/opencontrol/schemas) change?

models/components.go
```diff
@@ -91,19 +99,28 @@ func (components *Components) GetAll() map[string]*Component {
...
 		return ErrReadFile
 	}
 	err = yaml.Unmarshal(componentData, &component)
+	// Check the component version to give a better error before the generic "ErrControlSchema"
+	if component != nil && component.SchemaVersion != constants.DefaultFloat32Value {
+		versionErr := schema_tools.VerifyVersion(fileName, "component", component.SchemaVersion, constants.MinComponentYAMLVersion, constants.MaxComponentYAMLVersion)
+		if versionErr != nil {
+			return versionErr
+		}
+	}
```
afeld: Mind moving this into a function?

afeld:
The code looks good, but not sure that I totally understand the motivation for this change. Mind giving some explanation? Is there a corresponding issue?

fixtures/component_fixtures/EC2/component.yaml
jcscottiii: we will. which also made me think we should have a `lib` folder in that repository. 
for example:
```
schema/
-- lib/
---- go/
------v1_0/
--------component.go
------v2_0/
--------component.go
------v3_0/
--------component.go
```
Then we would ensure that the schema repository stays up to date because you would have to import instead of using this internal component.go

Thoughts?

jcscottiii:
Definitely forgot to create the issue. Created issue https://github.com/18F/compliance-toolkit/issues/25 here.

models/components.go
jcscottiii: done!

fixtures/component_fixtures/EC2/component.yaml
afeld: Putting the models in that repository, you mean? Opened https://github.com/opencontrol/schemas/issues/12 to discuss further.

docx/docx.go
```diff
@@ -29,59 +28,62 @@ func (config *Config) BuildDocx() error {
...
+				var narrativeText string
+				for _, section := range justification.SatisfiesData.Narrative {
+					// If section header exists, let's print it's corresponding text and not the header itself.
+					if section.Key == sectionKey {
+						narrativeText = fmt.Sprintf("%s%s\n", narrativeText, section.Text)
+						found = true
 					}
-				})
-			}
+				}
```
afeld: Can we make/use a `GetSection` method on the `Narrative`?

docx/docx.go
```diff
@@ -29,59 +28,62 @@ func (config *Config) BuildDocx() error {
...
+				// If we actually had a section, let's get the component and add the narrative
+				if found {
+					// Print out the component name.
+					text = fmt.Sprintf("%s%s\n%s", text, component.Name, narrativeText)
+				}
```
afeld: Should a section that's not found give an error, or a warning, or output ["not implemented"](https://github.com/opencontrol/compliance-masonry/issues/136)?

docx/docx_test.go
```diff
@@ -15,41 +15,53 @@ import (
...
-		config := docx.Config{
-			OpencontrolDir: filepath.Join("..", "fixtures", "opencontrol_fixtures"),
-			TemplatePath:   filepath.Join("..", "fixtures", "template_fixtures", "test.docx"),
-			ExportPath:     exportPath,
-		}
-		config.BuildDocx()
-		expectedDoc, _ := docTemp.GetTemplate(filepath.Join("..", "fixtures", "exports_fixtures", "output.docx"))
-		actualDoc, _ := docTemp.GetTemplate(exportPath)
-		assert.Equal(GinkgoT(), expectedDoc.Document.GetContent(), actualDoc.Document.GetContent())
+		It("load the template and build the final docx", func() {
```
afeld: `s/load/loads`

tools/version/compare.go
```diff
@@ -0,0 +1,17 @@
+package version
+
+import "github.com/opencontrol/compliance-masonry/tools/constants"
+
+// VerifyVersion will check if the version is compatible. If not, it will return a IncompatibleVersionError.
+func VerifyVersion(file string, fileType string, version, minVersion, maxVersion float32) error {
```
afeld: This is a lot of arguments...can we make a `SchemaFile` type or something?

tools/constants/constants.go
```diff
@@ -24,3 +24,20 @@ const (
...
+const (
+	// VersionNotNeeded is a place holder to indicate that the schema version does not need to be specified.
+	VersionNotNeeded float32 = -18.0
+	// MinComponentYAMLVersion is the minimum schema version for the component
+	// YAML supported by this masonry toolchain.
+	MinComponentYAMLVersion float32 = 3.0
+	// MaxComponentYAMLVersion is the minimum schema version for the component
+	// YAML supported by this masonry toolchain.
+	MaxComponentYAMLVersion float32 = VersionNotNeeded
+)
```
afeld: Maybe we should use major and minor versions as integers, and a string for the full version? My internal alarm went off about using floats where the comparison might be off by a small fraction, though it _shouldn't_ be a problem for us since we won't be doing math with them?

afeld:
## Questions

* This pull request got big! Is it too late to split out the schema version checking from the `narrative` changes?
* With these changes, is `narrative` as a simple string in the YAML still supported? Does every control need narrative sections, or is a simple string sufficient in some cases?

## TODOs

* [ ] Update the `getControl` instructions in the README
* [ ] Create a corresponding pull request on https://github.com/opencontrol/schemas

docx/docx.go
jcscottiii: This formatting is specific for docx. A similar function is in the gitbook here: https://github.com/opencontrol/compliance-masonry/pull/143/files#diff-53f404a533ed5d47e5f222aa94b2e428R20

Maybe in the future we can create an interface where all data that can be outputted to a template either gitbook or docx implements that interface e.g
```go
type TemplateData interface {
    FormatAllForDocx(standardKey, controlKey string) (string, error)
    FormatOneForDocx(standardKey, controlKey string, key string) (string, error)
    FormatAllForGitbook(standardKey, controlKey string) (string, error)
    FormatOneForGitbook(standardKey, controlKey string, key string) (string, error)
}
```

docx/docx.go
jcscottiii: I'm not sure and don't have a strong opinion towards either printing a warning or printing nothing lol.

The only thing I might leave outputting "not implemented" or other state to another template function because we do have that state in another variable in the `Satisfies` struct

docx/docx_test.go
jcscottiii: oops lol

docx/docx.go
afeld: Cool...maybe just split it out into a function above then? Just trying to make `FormatControlSection` a bit easier to follow.

tools/version/compare.go
jcscottiii: will fix

tools/constants/constants.go
jcscottiii: i agree, this isn't the best way period.

the opencontrol.yaml actually uses semver (and there's a nice library to handle this). i'm going to see if i can convert component.yaml to using semver without breaking all pre `v3.0` schemas.

docx/docx.go
jcscottiii: Will do!

docx/docx.go
jcscottiii: @mzia thoughts?

jcscottiii: just talked to @mzia. Just copying what he said here:
> I wouldn’t leave the tags mainly because that’s template data.  I’d leave it blank.  We should feed the word doc base on what’s in our yaml content.  We will be running it through the template to update, so when there is new content, the next ‘build’ will populate accordingly right?

mogul: I'd make it `No information available`.

jcscottiii:
@afeld should be ready to go

tools/version/compare.go
jcscottiii: Finally fixed here: https://github.com/opencontrol/compliance-masonry/pull/143/commits/43e45c4bdd17394a4c8dbfdba02a09c47f42f041

models/components.go
```diff
@@ -24,7 +28,93 @@ type Component struct {
...
+// (type =float).
+// To compensate for that, we have to hand roll our own UnmarshalYAML that can decide what to do for parsing
+// the older version of type float and converting it into semver. In addition, we will use this logic to parse strings
+// into semver.
+func (c *Component) UnmarshalYAML(unmarshal func(v interface{}) error) error {
+	// When we call "unmarshal" callback on an object, it will call that object's "UnmarshalYAML" if defined.
+	// Since we are currently in the implementation of Component's "UnmarshalYAML", when finally we call
+	// unmarshal again, if it's on type Component, we would end up in a recursive infinite loop.
+	// To prevent this, we create a separate type, called Alias.
+	type Alias Component
```
afeld: Questions:

* I think I skipped this chapter in Tour of Go 😄  This is like making a subclass, kinda?
* Does it matter that this is created within the method? Is it any more efficient to define it outside?

models/components.go
```diff
@@ -24,7 +28,93 @@ type Component struct {
...
+		// Schema Version started being documented with "2.0".
+		// We should be able to parse it for backwards compatibility.
+		// All future versioning should be in semver format already.
+		case 2.0:
+			ver = semver.MustParse("2.0.0")
+		// If not the older version, it needs to be in semver format, send an error.
+		default:
+			return componentLoadError{fmt.Sprintf(constants.ErrVersionNotInSemverFormatf, v)}
+
+		}
```
afeld: This seems like it could be made simpler, without the hard-coding:

```go
ver := fmt.Sprintf("%.1f.0", v)
ver // => "2.0.0"
```

afeld: Regardless, I'd move this into a `parseFloatVersion()` function.

models/components.go
```diff
@@ -24,7 +28,93 @@ type Component struct {
...
+	default:
+		return componentLoadError{fmt.Sprintf(constants.ErrMissingVersion)}
+	}
+	// Copy everything from the Alias back to the original component.
+	*c = (Component)(aux.Alias)
+
+	// Get the version
+	c.SchemaVersion = ver
+	return err
+}
```
afeld: This function is some super Go voodoo. You are a wizard 👏

afeld: Also, if this ends up being too brittle, it wouldn't be the end of the world to do a parse of the YAML file with the anonymous struct above (or even a regex) to find the version number, then do a normal unmarshalling with the appropriate versioned model.

models/components.go
```diff
@@ -24,7 +28,93 @@ type Component struct {
...
+
+// VerifySchemaCompatibility will check that the current component schema version is
+// compatible with the current masonry toolchain.
+func (c *Component) VerifySchemaCompatibility(fileName string) error {
+	if c != nil {
```
afeld: Could this function even be called if the pointer were `nil`?

models/components.go
```diff
@@ -91,19 +187,36 @@ func (components *Components) GetAll() map[string]*Component {
...
+	// If we don't have a user friendly error yet...
+	// Check the component version to give a better error before the generic "ErrControlSchema"
+	if versionErr := component.VerifySchemaCompatibility(fileName); versionErr != nil {
+		return versionErr
+	}
+
+	// If no specific errors were found, but a general error was found, return that.
 	if err != nil {
 		return ErrControlSchema
 	}
```
afeld: Isn't this now redundant?

tools/constants/constants.go
```diff
@@ -24,3 +26,37 @@ const (
 	// Components is the placeholder for the resource type of components
 	Components ResourceType = "Components"
 )
+
+// DefaultFloat32Value is the exported constant value for the default float32 value for an uninitialized float32
+// variable.
+const (
+	DefaultFloat32Value float32 = iota
+)
```
afeld: This doesn't seem to be used, unless I'm misunderstanding something...?

models/components_test.go
```diff
@@ -85,15 +148,28 @@ var componentTestErrors = []componentTestError{
 	{"", ErrComponentFileDNE},
 	// Check loading a component with a broken schema
 	{filepath.Join("..", "fixtures", "component_fixtures", "EC2BrokenControl"), ErrControlSchema},
+	//Check loading an older schema without a key.
+	{filepath.Join("..", "fixtures", "component_fixtures", "EC2BadVersion2_0"), version.NewIncompatibleVersionError(version.NewRequirements(filepath.Join("..", "fixtures", "component_fixtures", "EC2BadVersion2_0", "component.yaml"), "component", semver.MustParse("2.0.0"), constants.MinComponentYAMLVersion, constants.MaxComponentYAMLVersion))},
+	//Check loading an older schema with a key.
+	{filepath.Join("..", "fixtures", "component_fixtures", "EC2WithKeyBadVersion2_0"), version.NewIncompatibleVersionError(version.NewRequirements(filepath.Join("..", "fixtures", "component_fixtures", "EC2WithKeyBadVersion2_0", "component.yaml"), "component", semver.MustParse("2.0.0"), constants.MinComponentYAMLVersion, constants.MaxComponentYAMLVersion))},
+	// Check for versions not in semver format.
+	{filepath.Join("..", "fixtures", "component_fixtures", "EC2BadVersion_NotSemver"), componentLoadError{fmt.Sprintf(constants.ErrVersionNotInSemverFormatf, 3)}},
```
afeld: I don't think hard-coding the error strings is a terrible idea, to ensure that they are getting created as expected and there aren't any copy-and-paste errors.

tools/constants/constants.go
```diff
@@ -24,3 +26,37 @@ const (
...
+
+const (
+	ErrVersionNotInSemverFormatf = "Version %v is not in semver format"
+	ErrMissingVersion            = "Schema Version can not be found."
+)
+
+var (
+	// VersionNotNeeded is a place holder to indicate that the schema version does not need to be specified.
+	VersionNotNeeded semver.Version = semver.MustParse("0.0.0-NotNeeded")
```
afeld: This feels like a hack, and is confusing. Is there a way around it?

tools/constants/constants.go
```diff
@@ -24,3 +26,37 @@ const (
...
 )
+
+// DefaultFloat32Value is the exported constant value for the default float32 value for an uninitialized float32
+// variable.
+const (
+	DefaultFloat32Value float32 = iota
+)
+
+const (
+	ErrVersionNotInSemverFormatf = "Version %v is not in semver format"
```
afeld: These format string constants should be functions instead. Ditto below.

tools/version/compare.go
```diff
@@ -0,0 +1,36 @@
...
+		return nil
+	} else if r.version.GTE(r.minVersion) && r.maxVersion.EQ(constants.VersionNotNeeded) {
+		return nil
+	} else if r.minVersion.EQ(constants.VersionNotNeeded) && r.version.LTE(r.maxVersion) {
+		return nil
+	} else if r.version.GTE(r.minVersion) && r.version.LTE(r.maxVersion) {
+		return nil
+	}
+	return NewIncompatibleVersionError(r)
+}
```
afeld: Hmm, seems like `file` and `fileType` aren't actually used?

I would simplify this struct to only contain the `minVersion` and `maxVersion`, then pass the `version` as an argument to `VerifyVersion()`.

Not sure if it makes things that much simpler, but you might also consider using a `Range`:

```go
checkRequirements := semver.Range(func(v semver.Version) bool {
	return v.GTE(minVer) && v.LT(maxVer)
})
checkRequirements(r.version) // => true/false
```

jcscottiii:
A lot of the version stuff actually goes away in #151. I think i will make some new PRs

models/components.go
jcscottiii: dangerously enough. yeah https://play.golang.org/p/xsJhhDblf9

models/components.go
jcscottiii: The logic inside could be moved to another function but the go-yaml library is expecting your struct to have an `UnmarshalYAML`

so it could be
```go
func (s YourStruct) UnmarshalYAML(unmarshal func(v interface{}) error) error {
    somefunctioncall(unmarshal, c)
}
```

What do you have in mind?

models/components.go
jcscottiii: I moved it only allowing float for version `2.0`. To stop a user from continuing to use floats, if they pass `3.0` it will error out instead.

jcscottiii: Will move to separate function!

models/components.go
afeld: Well _that's_ annoying.

afeld: In that case, why here but not other methods?

models/components.go
afeld: I mean the `type` declaration specifically. Is this any more/less efficient than

```go
type Alias Component

func (c *Component) UnmarshalYAML(unmarshal func(v interface{}) error) error {
  // ...
}
```

?

jcscottiii: Ahh yeah, so it's basically a useless struct to other functions. I don't even want other functions in the same package access to this.

afeld:
Should we close this then, or are you thinking of rebasing on top of the other changes?

jcscottiii:
@afeld yeah let's close up shop and use the smaller PRs

PR #144, opened 2016-05-24
move docs file into its own package

afeld:
Hey look, I Go'd!

cool gopher

### DIFF
--- a/docs.go
+++ b/docs.go
@@ -33,7 +33,6 @@ func MakeGitbook(config gitbook.Config) []string {
 		return messages
 	}
 	if _, err := os.Stat(config.MarkdownPath); os.IsNotExist(err) {
-		markdownPath = ""
 		messages = append(messages, "Warning: markdown directory does not exist")
 	}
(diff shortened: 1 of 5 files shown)


jmcarp:
LGTM, although I don't have strong opinions about how big packages should be. Should `diff` and `get` also be split out? The [language recommendations](https://blog.golang.org/organizing-go-code) kind of equivocate on this:

> It is easy to just throw everything into a "grab bag" package, but this dilutes the meaning of the package name (as it must encompass a lot of functionality) and forces the users of small parts of the package to compile and link a lot of unrelated code.

> On the other hand, it is also easy to go overboard in splitting your code into small packages, in which case you will likely becomes bogged down in interface design, rather than just getting the job done.

afeld:
👍 Mind if I take care of that separately?

jmcarp:
Sure! I'm basically agnostic about whether that code lives in `main` or in separate packages, just wanted to understand the motivation.

dlapiduz:
LGTM

PR #145, opened 2016-05-24
move all non-main packages under single directory

afeld:
Builds on https://github.com/opencontrol/compliance-masonry/pull/144.

The top-level directory was getting a bit crowded, and while it doesn't seem uncommon to leave those package folders at the top level (e.g. [the Cloud Foundry CLI](https://github.com/cloudfoundry/cli)), I like it better when they are all under a single folder. This is akin to Java using a `src/` folder.

No other changes besides changing the imports to match.


jcscottiii:
Hmm. Not sure about putting these all under `packages`.
If we consolidate, the folders should have a purpose.
```
.
├── circleci
├── config
├── cmds
│   ├── inventory/
│   ├── docs/
    │   ├── docx/
    │   ├── gitbook/
│   ├── get/
├── fixtures
├── tools
├── vendor
```

The diff.go and diff tests will go into `inventory` and `inventory/` will be renamed `diff/` (since the command is diff)

The get.go will be moved to `cmds/get`

Also, did a quick check, the `testbed` folder is no longer needed / used.

afeld:
Ok, got most of the way through refactoring to the suggested format. Moving `diff.go` out of the `main` package turned out to be a bit tricky due to the shared data that's stored at the package level that it relies on, so am going to punt on that for now.

jcscottiii:
this looks good. just need to fix some paths for fixtures for the tests to pass

afeld:
Fixed!

PR #154, opened 2016-06-08
Fix false positive on CI tests for appveyor

jcscottiii:


appveyor.yml
```diff
@@ -19,4 +19,4 @@ install:
 build_script:
 - cmd: go build
 test_script:
-- cmd: FOR /F %%A IN ('glide novendor') DO go test -v %%A
\ No newline at end of file
+- cmd: FOR /F %%A IN ('glide novendor') DO go test -v %%A || exit /b 1
```
afeld: I think this can be even simpler with `go test -v $(glide nv)`.

jcscottiii: that syntax won't work for windows.

afeld: `/me shakes fist at Windows`

PR #155, opened 2016-06-08
Move to using interface for component, satisfies

jcscottiii:
This patch is the beginning of enabling multiple versions by
abstracting the logic to a common interface between versions

### DIFF
--- a/models/components.go
+++ b/models/components.go
@@ -8,11 +8,13 @@ import (
 	"sync"
 
 	"gopkg.in/yaml.v2"
+	"github.com/opencontrol/compliance-masonry/models/common"
+	"github.com/opencontrol/compliance-masonry/models/components/versions/base"
 )
 
 // Components struct is a thread-safe structure mapping for components
 type Components struct {
-	mapping map[string]*Component
+	mapping map[string]base.Component
 	sync.RWMutex
 }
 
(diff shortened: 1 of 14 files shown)


models/components/versions/base/component.go
```diff
@@ -0,0 +1,22 @@
...
+	GetReferences() *common.GeneralReferences
+	GetVersion() float32
+}
+
+type Satisfies interface {
+	GetStandardKey() string
+	GetControlKey() string
+	GetNarrative() string
+	GetCoveredBy() common.CoveredByList
+}
```
afeld: This probably warrants some developer-centric documentation (or at least a comment at the top of the file saying "we are using `interface`s to provide an abstraction between schema versions. See `some/path/` for the implementations of the different versions."

models/components.go
```diff
@@ -21,14 +23,44 @@ type Components struct {
...
+func (c Component) GetVerifications() *common.VerificationReferences {
+	return &c.Verifications
+}
+
+func (c Component) GetReferences() *common.GeneralReferences {
+	return &c.References
+}
+
+func (c Component) GetAllSatisfies() []base.Satisfies {
+	// Have to do manual conversion
```
afeld: Minor: I'd add "...to the more generic `interface` `type`."

jcscottiii: Will do

afeld:
Looks good overall! Mind adding tests to cover the lines that Codecov is complaining about?

models/components/versions/base/component.go
jcscottiii: done

PR #156, opened 2016-06-08
Separate version from common for component

jcscottiii:
This PR refactors the code so that particular versions of components will live in `models/components/versions/X_Y_Z`

In addition, this patch introduces an intermediate parser which determines the schema version then will detect which type of schema to then continue trying to unmarshal into


jcscottiii:
Depends on #155

jcscottiii:
@afeld

afeld:
Moved to https://github.com/opencontrol/compliance-masonry/pull/157 to change the base branch to `master`.

PR #157, opened 2016-06-08
Separate version from common for component, Take II

afeld:
Moved from https://github.com/opencontrol/compliance-masonry/pull/156.

> This PR refactors the code so that particular versions of components will live in `models/components/versions/X_Y_Z`
>
> In addition, this patch introduces an intermediate parser which determines the schema version then will detect which type of schema to then continue trying to unmarshal into


jcscottiii:
There's one thing that I'm addressing in the next PR, the formatted string as consts. One is here. the next PR it will be gone.

models/components/components.go
```diff
@@ -0,0 +1,61 @@
...
+// Add adds a new component to the component map
+func (components *Components) Add(component base.Component) {
+	components.Lock()
+	components.mapping[component.GetKey()] = component
+	components.Unlock()
+}
+
+// Get retrieves a new component from the component map
+func (components *Components) Get(key string) base.Component {
+	components.Lock()
```
afeld: Can this be an [`RLock()`](https://golang.org/pkg/sync/#RWMutex.RLock)?

models/components/versions/base/component.go
```diff
@@ -19,4 +19,9 @@ type Satisfies interface {
 	GetControlKey() string
 	GetNarrative() string
 	GetCoveredBy() common.CoveredByList
+}
+
+// Base is the common struct that all component schemas must have.
```
afeld: Not a blocker, but I wonder if there's a Go-type-system-friendly way to enforce this.

models/opencontrol.go
```diff
@@ -110,3 +110,29 @@ func (openControl *OpenControl) LoadStandards(standardsDir string) error {
...
+		return err
+	}
+	if component.GetKey() == "" {
+		component.SetKey(getKey(componentDir))
+	}
+	if openControl.Components.CompareAndAdd(component) {
+		openControl.Justifications.LoadMappings(component)
+	}
+	return nil
+}
```
afeld: Mind splitting up this method?

models/components/versions/base/component.go
jcscottiii: Ahh actually I need to fix this comment.

In a previous iteration of this, I was embedding the `Base` in all the versions of `Component` But here and in the future PRs, I won't be anymore. I don't need to / can't anymore since I'm doing that Unmarshal trick in #158 anyhow.

Needs to read `Base is the bare minimum that every Component will have and is used to find the semver Version`.

models/components/components.go
jcscottiii: Yep!

models/opencontrol.go
jcscottiii: wrapped the file system functions into one function.

models/components/components.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"log"
+	"sync"
+
+	"github.com/opencontrol/compliance-masonry/models/components/versions/base"
+)
+
+// Components struct is a thread-safe structure mapping for components
+type Components struct {
+	mapping map[string]base.Component
```
afeld: Not a blocker (since this existed before this pull request), but can we call this something more informative, like `componentsByKey`?

models/components/components.go
```diff
@@ -0,0 +1,61 @@
...
+// This function is thread-safe.
+func (components *Components) CompareAndAdd(component base.Component) bool {
+	components.Lock()
+	defer components.Unlock()
+	added := false
+	if _, exists := components.mapping[component.GetKey()]; !exists {
+		components.mapping[component.GetKey()] = component
+		added = true
+	} else {
+		log.Fatalf("Component: %s exisits!\n", component.GetKey())
```
afeld: Not a blocker (same reason), but wouldn't we want to return an `error` here? (I feel like I asked this before...sorry if I forgot the response!)

PR #160, opened 2016-06-09
V3 0 0 component

jcscottiii:
This depends on #159 
This patch introduces the following fields to the component.yaml: `parameters` `responsible_role`
This patch modifies the following fields: `narrative`

`narrative` now contains two fields: `key` and `text`. the `key` field is optional
`parameters` now contains two fields: `key` and `text`. both fields are required if used.
`responsible_role` is a single string.

These new fields are what make the 3.0.0 version schema of component.yaml

This PR does not change currently how the information is rendered to gitbook or docx. That will come in a subsequent PR.

An example schema for 3.0.0 is as follows:
```yaml
documentation_complete: false
name: Amazon Elastic Compute Cloud
key: EC2
references:
- name: Reference
  path: http://VerificationURL.com
  type: URL
satisfies:
- control_key: CM-2
  covered_by:
  - verification_key: EC2_Verification_1
  - component_key: UAA
    system_key: CloudFoundry
    verification_key: UAA_Verification_1
  implementation_status: partial
  narrative:
    - key: "a"
      text: "Justification in narrative form A for CM-2"
    - key: "b"
      text: "Justification in narrative form B for CM-2"
  standard_key: NIST-800-53
- control_key: 1.1
  covered_by:
  - verification_key: EC2_Verification_1
  - component_key: UAA
    system_key: CloudFoundry
    verification_key: UAA_Verification_1
  implementation_status: partial
  parameters:
    - key: "a"
      text: "Parameter A for 1.1"
    - key: "b"
      text: "Parameter B for 1.1"
  narrative:
    - key: "a"
      text: "Justification in narrative form A for 1.1"
    - key: "b"
      text: "Justification in narrative form B for 1.1"
  standard_key: PCI-DSS-MAY-2015
- control_key: 2.1
  covered_by: []
  implementation_status: partial
  narrative:
    - text: "Justification in narrative form for 2.1"
  standard_key: PCI-DSS-MAY-2015
responsible_role: "AWS Staff"
schema_version: 3.0.0
verifications:
- key: EC2_Verification_2
  name: EC2 Governor 2
  path: artifact-ec2-1.png
  type: Image
- key: EC2_Verification_1
  name: EC2 Verification 1
  path: http://VerificationURL.com
  type: URL

```

PR #161, opened 2016-06-09
V3 0 0 component rendering into gitbook and docx formats

jcscottiii:
This patch will introduce ways to render the version 3.0.0 components which were introduced in #160.

For Gitbook:

- The Narrative Key is printed as a header
- The Narrative Text is printed afterwards under the narrative key.
- Responsible Role is printed after the Component Name.

For Docx:

There are new templating functions:

- `{{getResponsibleRole "standard" "control"}}` - Will get the responsible role.
- `{{getAllControlSections “standard” "control”}}` - Will get all the narratives for a control
- `{{getControlSection “standard” “control” “section key”}}` - Will get the narrative for that given section of that control.
- {{getParameter “standard” “control” “parameter key”}}` - Will get the parameter for that given parameter key of that control.

Docx Template:
image

Docx Filled Template:
image


docx/docx.go
```diff
@@ -23,66 +24,156 @@ type OpenControlDocx struct {
...
+	}
 	docTemplate.AddFunctions(funcMap)
 	docTemplate.Parse()
 	docTemplate.Execute(config.ExportPath, nil)
 	return err
 }
 
-// FormatControl returns a control formatted for docx
-func (openControl *OpenControlDocx) FormatControl(standardControl string) string {
+type componentInfoType int
```
afeld: Is there a point to making the sub-type here? (legitimately asking)

docx/docx.go
```diff
@@ -23,66 +24,156 @@ type OpenControlDocx struct {
...
+
+			// If section header exists, let's print it's corresponding text and not the header itself.
+			if specifiedSections.Has(section.GetKey()) {
+				text = fmt.Sprintf("%s%s\n", text, section.GetText())
+				foundText = true
+			}
+		}
+	}
+	return text, foundText
+}
```
afeld: This function is doing a lot...can we split it up? Example functions that could come out of it:

* `getCombinedControlInfo()`
* `getSectionInfo()`
* `getNarrative()`

Also, not sure that `foundText` is necessary, but it's not hurting anything I suppose.

docx/docx.go
```diff
@@ -23,66 +24,156 @@ type OpenControlDocx struct {
...
+
+	for _, parameter := range justification.SatisfiesData.GetParameters() {
+		// If section header exists, let's print it's corresponding text and not the header itself.
+		if specifiedSections.Has(parameter.GetKey()){
+			text = fmt.Sprintf("%s%s\n", text, parameter.GetText())
+			foundText = true
+		}
+	}
+	return text, foundText
+}
```
afeld: Mind extracting a `getParameter()` function?

docx/docx.go
```diff
@@ -23,66 +24,156 @@ type OpenControlDocx struct {
...
-						verification := component.GetVerifications().Get(coveredBy.VerificationKey)
-						text += fmt.Sprintf("- %s %s  \n", verification.Name, verification.Path)
-					}
-				})
-			}
+				if found {
+					text = fmt.Sprintf("%s%s", text, specificText)
+				} else {
+					text = fmt.Sprintf("%s%s%s\n", text, specificText, constants.WarningNoInformationAvailable)
+				}
```
afeld: Can we extract this to a method?

gitbook/gitbookCertification.go
```diff
@@ -13,10 +13,31 @@ func (openControl *OpenControlGitBook) exportControl(control *ControlGitbook) (s
...
+					text = fmt.Sprintf("%sNo narrative found for the combination of standard %s and control %s\n", text, control.standardKey, control.controlKey)
+					return
+				}
+
+				for _, narrative := range justification.SatisfiesData.GetNarratives() {
+					if narrative.GetKey() != "" {
+						text = fmt.Sprintf("%s\n##### %s\n", text, narrative.GetKey())
+					}
+					text = fmt.Sprintf("%s%s\n", text, narrative.GetText())
+				}
```
afeld: Can we move everything/most from this callback into a function?

afeld: Ditto with the outer callback. Basically, break `exportControl()` into smaller pieces.

docx/docx.go
jcscottiii: yeah. I want so create some type of enumeration of what type of data i want going into the common function `getComponentText`. i can't pass a int by itself, only through those value declared by my new type.

afeld:
Merge from `master` (so the PR updates) when ready!

docx/docx_test.go
```diff
@@ -1,55 +1,104 @@
-package docx_test
+package docx
```
afeld: Is this right?

jcscottiii: yeah. So when you split out your test to the `packagename_test` (in this case `docx_test`) package, you can only test exported functions.

However, there was some edge case here https://github.com/opencontrol/compliance-masonry/pull/161/files#diff-13ce5f19e770df462c7202a0913535e5R80
that I didn't have a fixture for. So i needed access to an internal method so i have to move the test to the `packagename` (in this case `docx`) package.

afeld: I'd add an inline comment about that, in that case. After that, 🚢 !

PR #181, opened 2016-06-23
Needed a useful error. Pardon the bad go

pburkholder:
We'll probably want to reject this for a standard approach to emitting useful error message. Anyhow my first PR in GoLang (gotta start someplace).


models/opencontrol.go
```diff
@@ -83,7 +83,7 @@ func (openControl *OpenControl) LoadComponents(directory string) []error {
 	var wg sync.WaitGroup
 	componentsDir, err := ioutil.ReadDir(directory)
 	if err != nil {
-		return []error{ErrReadDir}
+		return []error{errors.New("Error: Unable to read the directory "+directory)}
```
afeld: We might be able to simply do

```go
return []error{err}
```

I'd do that for the other occurrences of `ErrReadDir` too, and get rid of that error completely for now.

afeld:
> We'll probably want to reject this for a standard approach to emitting useful error message.

The problem in this case is that we were being too clever in trying to pass a custom error message rather than just surfacing the pre-existing error. I think that's fine for now, until we want to get really crafty with the error messages.

> my first PR in GoLang

🎉

afeld:
Also, if you're feeling bold, mind adding a test to ensure that the directory is reported? It will go into `models/opencontrol_test.go`, and can look like a variation of [this](https://github.com/opencontrol/compliance-masonry/blob/9b873de27bf726fb4d15fff19cafb281e9222664/models/opencontrol_test.go#L58-L63), but checking the returned `error`.

models/opencontrol.go
pburkholder: I forgot about this while I was writing tests. It works with the string, now, but now that I have tests working we could go this route too.

pburkholder:
ErrReadDir is gone and tests are in place. Good to go?

models/opencontrol.go
afeld: We might need a more holistic approach, but I'm good with this for now.

afeld:
Thanks so much!

PR #190, opened 2016-07-01
Update Readme.md to add `go build` step

neilernst:
At least for me, just doing `go get` did not produce the `compliance-masonry` binary mentioned in the subsequent step. This adds to the docs the build step that produces the binary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,13 @@ Compliance Masonry is a command-line interface (CLI) that allows users to constr
     go get github.com/opencontrol/compliance-masonry
     ```
 
+1. Build the tool
+
+    ```bash
+    cd $GOPATH/src/github.com/opencontrol/compliance-masonry/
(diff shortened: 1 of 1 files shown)


afeld:
Hmm, per [the docs](https://golang.org/cmd/go/#hdr-Download_and_install_packages_and_dependencies) (and my experience),

> Get downloads and installs the packages named by the import paths, along with their dependencies.

First, are you seeing `compliance-masonry` in `$GOPATH/bin`? Second, is `$GOPATH/bin` in your `PATH`?

jcscottiii:
+1 what @afeld said about it being in `$GOPATH/bin`

neilernst:
Sorry, you're right. I was using tab-completion and zsh hadn't updated with the GO binary, so I assumed it wasn't there.

PR #194, opened 2016-07-12
remove schema information for the opencontrol.yaml

afeld:
All of the other information about the file schemas are in the schemas repository, so removing the opencontrol.yaml information here to be consistent. Depends on https://github.com/opencontrol/schemas/pull/36.


jcscottiii:
looks good. will merge when opencontrol/schemas#36 is merged

PR #201, opened 2016-07-20
fix reference to supported schema versions

afeld:
/cc @jbarnicle @jcscottiii

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ Compliance Masonry examples in the wild (in order of completeness):
 
 ## Documentation Format
 
-Compliance Masonry uses the [OpenControl v2 Schema](https://github.com/opencontrol/schemas).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -167,7 +167,7 @@ Compliance Masonry examples in the wild (in order of completeness):
 
 ## Documentation Format
 
-Compliance Masonry uses the [OpenControl v2 Schema](https://github.com/opencontrol/schemas).
+Compliance Masonry uses the [OpenControl Schema](https://github.com/opencontrol/schemas), `v2.0.0` through `v3.1.0`.
```
jcscottiii: Maybe we want to just put the link and not specify a version. opencontrol.yaml uses only 1.0.0 and then component has it's own range, could get confusing to try to specify here.

afeld: Hmm, good point. Do you think it's useful to call out the supported schema versions at all?

jcscottiii: I do think we need to have a table
just a mapping of yaml filename  and supported versions

PR #205, opened 2016-07-25
Rename models package to lib package

### LINKED ISSUE #204: rename models to lib
Feeling inspired by https://peter.bourgon.org/go-best-practices-2016/#repository-structure
To make it clear in terms which packages should be imported by other projects, we should have a lib folder. I think `models` is essentially that.

After we do that, we need to go through the code and see what's exported vs private inside `lib`
The interfaces should be exported. The concrete structs should be private.

jcscottiii:
Closes #204 

This PR just renames the `models` package to the `lib` package and fixes all references.

Related to https://github.com/opencontrol/fedramp-templater/issues/24


jcscottiii:
Ready for review

PR #206, opened 2016-07-25
Rename opencontrol to workspace

jcscottiii:
This PR renames opencontrol to `LocalWorkspace`

Builds on #205 

Solves part of #202

PR #207, opened 2016-07-25
Rename config to opencontrol

### LINKED ISSUE #203: move config into lib and rename config to opencontrol
I did this and it's been nagging me for awhile when I developed `get` in my own silo without knowledge of the rest of the code. 

`config` is really `opencontrol.yaml`. It should be with all the other yaml file representations in `models`. Once #202 is done, opencontrol will be available and we can move config/opencontrol into `models`

It should essentially have the same structure as component.

jcscottiii:
Closes #203 

Builds on #206 

- Created a `common` package inside lib and moved the contents of `config/common` in there. 
  - Other packages were using it, so it makes sense to move it to a broader location and not specifically in `config`.
- Renamed `lib/config` to `lib/opencontrol`
- Renamed `ResourceGetter` to `Getter` inside the package `resources`
  - Now, it reads as `resources.Getter` instead of `resources.ResourceGetter`
- Renamed `ConfigWorker` to `Worker`.
- Renamed the interface`BaseSchema` to `OpenControl` 
  - Each version of the opencontrol.yaml will implement the functions of the `OpenControl` interface.


jcscottiii:
Ready to review!

Issue #153, opened 2016-06-07
As a Masonry user, I want the ability to use a custom formatter


afeld:
In order to have more control over how output is generated (implementation language, templating system, target-specific logic), we want the formatters in Compliance Masonry to be separate codebases that are called at the right time.

**Acceptance Criteria**

- [ ] There's a way to specify a custom output formatter when invoking compliance-masonry
- [ ] There's a "null" formatter that doesn't do much other than just show that custom formatters are possible

**Implementation notes**

- Have to work out the API/CLI between CM and the formatters
- Lots of refactoring work to pull out related tests, etc.

Original post below.
---

There's a decent amount of code in Masonry that's specific to docx or gitbook output...thinking it would be great to get that formatter-specific functionality out of Masonry itself. This could mean a plugin-esque architecture, or thinking of Masonry as a library and making it easy to create wrapper CLIs that are more specific to the use case. Thinking of the [fedramp-templater](https://github.com/opencontrol/fedramp-templater) specifically, where some FedRAMP, where the "simple" docx templating might not be sufficient.

joshuamckenty:
GitBook itself has a “formatter” plugin architecture, and the ability to generate a number of different output formats. Any reason we wouldn’t put this in there?

geramirez:
I definitely like the idea of distributing masonry's functionality as the project grows. Just to throw another thought in the mix; the [cf cli has a great way of allowing plugins](https://github.com/cloudfoundry/cli/tree/master/plugin_examples).

afeld:
> GitBook itself has a “formatter” plugin architecture

I found https://developer.gitbook.com/plugins/index.html, but am not seeing anything about adding output formats...have any links?

> Any reason we wouldn’t put this in there?

If it works, fine by me!

> the cf cli has a great way of allowing plugins

Was just going to ask if that functionality is reusable/extractable, but a quick search brought up a few packages that are designed for just that:

* https://golang.org/pkg/net/rpc/
* https://github.com/hashicorp/go-plugin
* https://github.com/natefinch/pie

Sounds like a good way to go to me!

pburkholder:
I'm not sure how I feel about this. The current compliance-masonry workflow is hard enough to follow, and has enough inconsistencies, that adding another layer of abstraction with plugins feels premature.

My gut says I'd like to see consistent generation of content for `gitbook serve` and `gitbook pdf` with the current flow before adding in a plugin architecture. My gut has an open mind though.

[...]

PR #210, opened 2016-07-28
remove docx templating logic

afeld:
As stated in #153, I don't think there's a lot of value in supporting/using Go-style template tags in Word documents, so let's take the logic out of the core tool. If there's interest in continuing to use the template tags in Word docs, we can move this logic into its own project.


glide.lock
```diff
@@ -40,58 +34,49 @@ imports:
...
   - unix
 - name: gopkg.in/fatih/set.v0
-  version: 27c40922c40b43fe04554d8223a402af3ea333f3
+  version: 26ad09bd764b8f3b6a2489753642774559fe049c
 - name: gopkg.in/yaml.v2
-  version: a83829b6f1293c91addabc89d0571c246397bbf4
+  version: 26ad09bd764b8f3b6a2489753642774559fe049c
 - name: vbom.ml/util
-  version: db5cfe13f5cc80a4990d98e2e1b0707a4d1a5394
+  version: 26ad09bd764b8f3b6a2489753642774559fe049c
```
afeld: Looks like `glide rm github.com/opencontrol/doc-template` corrected the `version`s to reflect what's in `vendor/`, rather than what's the latest in the respective repositories. My bad from earlier!

PR #211, opened 2016-07-28
add Code Climate configuration

### LINKED ISSUE #162: As a Masonry developer, I want static analysis results in my pull requests
Having the [Go Report Card](https://goreportcard.com/report/github.com/opencontrol/compliance-masonry) badge in the README is great, but a) it seems like it needs to be refreshed manually, and b) we don't get notified about new issues. We should set up [pull request integration through Code Climate](https://docs.codeclimate.com/docs/github#pull-requests), or similar.

afeld:
Closes #162.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,18 @@
+engines:
+  fixme:
+    enabled: true
+  gofmt:
+    enabled: true
+  golint:
+    enabled: true
+    exclude_paths:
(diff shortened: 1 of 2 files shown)


afeld:
Note the Code Climate failure is expected, since the config is new. It will only fail for newly-introduced problems in pull requests going forward.

https://docs.codeclimate.com/docs/github#pull-requests

PR #212, opened 2016-08-01
Add interfaces for workspace, other yamls.

jcscottiii:
To enable inter-operability between different versions of different things like standards and certifications (opencontrol.yaml and component.yaml already has it) and easy mocking, this patch introduces the use of more interfaces. Also, this patch introduces workspace interface which enables for external plugins to easily test by just using the interface to mock instead of worrying about fixtures.

## Major Additions:
- **Added example plugin, plugin tests with mocks and README.**
- Added interfaces for  `Certifications`, `Control`, `Standard`, `Workspace` to the `common` package
- Moved interfaces `Component`, `Satisfies`, and `Section` to the `common` package
-  Added importable mocks for `Component`, `Satisfies`, `Certifications`, `Control`, `Section`, `Standard`, `Workspace`
- Moved `Standard` to it's own package for future versioning and loading by versions.
  - A lot of plumbing is still needed here like adding a special UnmarshalYAML but the whole point of moving it was to move all yaml files into their own package, out of the base `lib` package.
- Moved `Certification` to it's own package for future versioning and loading by versions.
  - A lot of plumbing is still needed here like adding a special UnmarshalYAML but the whole point of moving it was to move all yaml files into their own package, out of the base `lib` package

## Other changes:
- Removed the use of callbacks
  - This will help with mocking
- Made the maps `Standards` and `Components` and their methods private.
  - This makes it more clear for people who import the lib package, which things they can use. If something is private, they won't see it.
- Fixed a lot of golint problems
  - When I moved files around, the old golint problems became new again so I had to fix them.

## Actual logic changes:
- For versions 2.0.0, and 3.0.0 of component, `GetImplementationStatuses` returned an empty array. It should return the implementation status if it's there.
 - Can probably be moved out of this PR but it's a small change compared to the rest.

For future PR:
- Add `verifications` to `common` package.
  - This will allow for `workspace` to also be added to `common` without creating a circular dependency chain.

# Excerpt from `lib/README.md`:
## Lib Package

This package (and the sub packages) are the only packages that should be imported by external projects. This lib package is very helpful for those writing plugins to extend the functionality of Masonry.

### `Workspace`
Workspace is the starting point for all the information available to plugins. For those using Go plugins, using the interface methods are all you need. It represents all the information collected into the `opencontrols` folder after running the `get` command. 

#### YAML files
The yaml files are represented by:

- `Components`
- `Certifications`
- `Standards`

#### Result Data
`Justifications` is a data structure that is not represented in yaml but rather a post-processed map of data to help quickly getting component data for a particular control name - standard name combination.
 

### Plugin Developer Guide
Developers should not have to worry about loading real data / workspaces for their unit tests (they should for integration tests).

There is an example of developing your Go plugin and tests in `exampleplugin/example.go` and `exampleplugin/example_test.go` respectively.


jcscottiii:
@afeld yeppp.. I did it again. this is a mega elephant patch. A lot of it is moving files around though. Details above. This PR was one of those that started off one way then turned 90 degrees to something else so it was hard to break it up. Now that it's done, I can do that now. Let me know and I'll start to break this up later. Not a huge priority now so I'll get to it later probably.

exampleplugin/example_test.go
```diff
@@ -0,0 +1,33 @@
+package main
+
+import (
+	"testing"
+	"github.com/opencontrol/compliance-masonry/lib/mocks"
+	commonMocks"github.com/opencontrol/compliance-masonry/lib/common/mocks"
```
afeld: Needs a space (right?)

exampleplugin/example.go
```diff
@@ -0,0 +1,23 @@
...
+	selectJustifications := p.GetJustification("standard", "control")
+	if len(selectJustifications) == 0 {
+		return "no data"
+	}
+	return selectJustifications[0].SatisfiesData.GetImplementationStatus()
+}
+
+func main() {
+	workspace, _ := lib.LoadData("sample opencontrol directory", "cert path")
+	fmt.Println(simpleDataExtractAndFormat(plugin{workspace}))
```
afeld: Minor: I'd split this to two lines so it's clearer what the return value of `simpleDataExtractAndFormat()` is.

exampleplugin/example_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+	// create mock workspace
+	ws := new(mocks.Workspace)
+	ws.On("GetJustification", "standard", "control").Return(lib.Verifications{})
+	// test function expecting "no data"
+	p := plugin{ws}
+	data := simpleDataExtractAndFormat(p)
+	assert.Equal(t, data, "no data")
+
+	// Test the case when there is data.
```
afeld: This should be a separate test.

lib/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+## `Workspace`
+Workspace is the starting point for all the information available to plugins. For those using Go plugins, using the interface methods are all you need. It represents all the information collected into the `opencontrols` folder after running the `get` command. 
+
+### YAML files
+The yaml files are represented by:
+
+- `Components`
+- `Certifications`
+- `Standards`
```
afeld: Maybe link to https://github.com/opencontrol/schemas for more information?

lib/certifications/certifications_test.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+		// Check if loaded certification has the expected number of standards
+		if len(actual.GetStandards()) != example.expectedStandards {
+			t.Errorf("Expected %d, Actual: %d", example.expectedStandards, len(actual.GetStandards()))
+		}
+		// Get the length of the control by using the GetSortedData method
+		totalControls := 0
+		standardKeys := actual.GetSortedStandards()
+		for _, standardKey := range standardKeys {
+			totalControls += len(actual.GetStandards()[standardKey].GetSortedControls())
```
afeld: Could `GetStandard()` be used here and the other places you're doing `.GetStandards()[standardKey]`?

lib/certifications/versions/1_0_0/certification.go
```diff
@@ -0,0 +1,39 @@
...
+	"sort"
+	"vbom.ml/util/sortorder"
+	v1standards "github.com/opencontrol/compliance-masonry/lib/standards/versions/1_0_0"
+	"github.com/opencontrol/compliance-masonry/lib/common"
+)
+
+// Certification struct is a collection of specific standards and controls
+// Schema info: https://github.com/opencontrol/schemas#certifications
+type Certification struct {
+	Key       string              `yaml:"name" json:"name"`
```
afeld: Should we be consistent about the YAML and struct field names?

lib/common/component.go
```diff
@@ -0,0 +1,76 @@
...
+//
+// GetVerifications get all the verifications.
+//
+// GetReferences get all the references.
+//
+// GetVersion returns the version
+//
+// SetVersion sets the version for the component.
+//
+// GetResponsibleRole gets the responsible party / role for the component.
```
afeld: Is it typical to have this documentation up here rather than above each individual function?

lib/common/mocks/Certification.go
```diff
@@ -0,0 +1,55 @@
+package mocks
+
+import "github.com/opencontrol/compliance-masonry/lib/common"
+import "github.com/stretchr/testify/mock"
+
+// Certification is an autogenerated mock type for the Certification type
```
afeld: Is this file auto-generated?

lib/components/component.go
```diff
@@ -1,45 +1,45 @@
...
+	component, err = parseComponent(componentData,fileName)
+	if err != nil {
+		return nil, err
+	}
+	// Ensure we have a key for the component.
+	if component.GetKey() == "" {
+		component.SetKey(getKey(path))
+	}
+	return component, nil
 }
```
afeld: Might want to split this up into a few smaller functions.

lib/components_test.go
```diff
@@ -0,0 +1,39 @@
+package lib
```
afeld: Ideally this would be the `lib_test` package, right? Or is that only when using Ginkgo?

lib/workspace_test.go
```diff
@@ -68,24 +44,26 @@ var loadDataTests = []loadDataTest{
...
 		if actualStandardsNum != example.expectedStandardsNum {
 			t.Errorf("Expected: `%d`, Actual: `%d`", example.expectedComponents, actualComponentNum)
 		}
+		/*
 		// Check the number of justifications
 		actualJustificationNum := len(actual.Justifications.mapping)
 		if actualJustificationNum != example.expectedJustificationNum {
 			t.Errorf("Expected: `%d`, Actual: `%d`", example.expectedComponents, actualComponentNum)
 		}
+		*/
```
afeld: Did you mean to leave this commented out?

afeld:
A few pieces of general feedback:

* Wow, yeah, this was massive...let's not do that again :grimacing:
* Love the example plugin and its documentation! How about we split its addition out to its own pull request?
* Not sure that I understand the `common` package...maybe it could have a more descriptive name?

Sorry for the slow turnaround on the review, by the way!

> For versions 2.0.0, and 3.0.0 of component, `GetImplementationStatuses` returned an empty array. It should return the implementation status if it's there. Can probably be moved out of this PR

If it's not hard, that would be good!

lib/common/component.go
jcscottiii: Only for interfaces.

https://www.reddit.com/r/golang/comments/2ynbfm/idiomatic_go_should_interface_methods_have/

lib/components_test.go
jcscottiii: we can do either usually. it comes down to whether we want to do blackbox (only access to the exported functions of `lib`) vs whitebox (access to public and private functions of `lib`)  testing. in this case, i need to be in the `lib` package because i needed to create the `components` map via `newComponents`

lib/workspace_test.go
jcscottiii: whoops lol

lib/certifications/versions/1_0_0/certification.go
jcscottiii: I saw this when I was moving around the code. Ideally, we should to prevent confusion. Will make a separate PR for this.

lib/components_test.go
afeld: Having them in the same package means the test files get included in the final build, right? Not a _huge_ problem, but this could make the resulting binary larger than it needs to be.

lib/certifications/certifications_test.go
jcscottiii: you're right!

Side note:
Ideally, I would want to get rid of the plural getters (e.g. `GetComponents`, `GetStandards`) because you have to return a container (in this case a map; map[string]Standard) which assumes what the key type is (in this case a string). Just wasn't sure

lib/common/mocks/Certification.go
jcscottiii: yeah via [mockery](https://github.com/vektra/mockery). Probably should put something in the development documentation about how to generate the mocks.

lib/components/component.go
jcscottiii: Will do!

exampleplugin/example_test.go
jcscottiii: whoops! will fix

exampleplugin/example.go
jcscottiii: will do!

exampleplugin/example_test.go
jcscottiii: Will do!

lib/README.md
jcscottiii: Will do!

jcscottiii:
> Not sure that I understand the `common` package...maybe it could have a more descriptive name?

Yeah, I agree. There are some things like `Component` and `Standard` there so it can read as `common.Component` and the `Workspace` is supposed to be able to return these common Components and common Standards (regardless of which version they are) but there were some existing things like `Entry` that don't belong in the same package. Will look into moving out some of that out.

lib/components_test.go
jcscottiii: This comment made me nervous because everyone likes lots of tests lol. I just experimented with this. While the test is in the same package as the code, `go build` does not package *_test.go files into the final binary. The package just is for scope when running `go test`

afeld: Thanks for checking!

afeld:
@jcscottiii Are you planning to rebase this one (sooner or later) or should we just close it?

afeld:
Bump!

jcscottiii:
Let's close this. The other recent PRs have been just parts of this one.
The only PRs yet to be made are 1) generate the mocks and 2) example plugin.

jcscottiii:
Split out to #213 #214 #215 #216 #217 #226 #227 #228

PR #213, opened 2016-08-19
Move Components interface to common package

jcscottiii:
This patch is just https://github.com/opencontrol/compliance-masonry/pull/212 broken into smaller pieces

This patch moves the components interface to the common package
Also, it get rids of the GetAndApply function that passes a callback
Also, it adds the mock for the component
  - The Component mock needed to be added in this PR because of the tests coverage was complaining due to coverage lowering and the mock helped with making tests

This patch does not unexport all the functions because the workspace
interface has not been created yet. Will come in another PR
A future PR will introduce instructions on how to generate the mocks. 
  - Right now, the tool mockery will error. There's a PR to fix it https://github.com/vektra/mockery/pull/95
  - However, it seems like it won't be merged anytime soon so a workaround and cleanse of the common package will happen later.


jcscottiii:
@afeld this is ready

lib/components.go
```diff
@@ -0,0 +1,76 @@
...
+	return components.mapping[key]
+}
+
+// CompareAndAdd compares to see if the component exists in the map. If not, it adds the component.
+// This function is thread-safe.
+func (components *componentsMap) CompareAndAdd(component common.Component) bool {
+	components.Lock()
+	defer components.Unlock()
+	added := false
+	if _, exists := components.mapping[component.GetKey()]; !exists {
```
afeld: While I know this isn't unusual for Go, I'm personally not a fan of cramming assignment plus a check into a conditional. I favor one thing per line.

lib/components.go
```diff
@@ -0,0 +1,76 @@
...
+		added = true
+	}
+	return added
+}
+
+// GetAll retrieves all the components without giving directly to the map.
+func (components *componentsMap) GetAll() []common.Component {
+	components.RLock()
+	defer components.RUnlock()
+	c := make([]common.Component, len(components.mapping))
```
afeld: How about calling this variable `results`?

jcscottiii: will do!

lib/components.go
jcscottiii: Done!

jcscottiii:
@afeld this should have touched on everything. and threw in some extra tests too lol

PR #214, opened 2016-08-19
Move standards interface

jcscottiii:
Builds on #213 

Breakout of #212 

Adds standard and control interfaces and puts it into the common interface.
Adds a mock for standard (was needed for testing)


jcscottiii:
@afeld this should be good after #213 is merged

PR #215, opened 2016-08-19
Move certifications to common

jcscottiii:
Add interface for certification to common.
Addresses https://github.com/opencontrol/compliance-masonry/pull/212/files#r74363210 by using `GetStandard` instead of `GetStandards`

A future PR will include the mock for it once the `common` package is cleaned up as talked about in #213

PR #216, opened 2016-08-19
Move justification to result package

jcscottiii:
Builds on #215 

This package is a location of all the data that has been added from all
the data loaded in and should be separated from the regular data in lib/


jcscottiii:
@afeld this is done

afeld:
:shipit:

PR #217, opened 2016-08-22
Refactor opencontrol interface and clean common part 1.

jcscottiii:
Builds on #216 

This was necessary to clean up `lib/common` package.

The `lib/common` package had both interfaces and structs for the YAMLs.
This led to weird import paths to get around circular dependencies.
This patch is just mostly moving stuff around.

- Renamed `Entry` to `RemoteSource` in the `lib/common` package
  - It was very vague what an `Entry` was. It really was a container of information to some remote (re)source
  - RemoteSource is now an interface, instead of a struct of yaml details
- The implementation of `RemoteSource` for the v1.0.0 of the `OpenControl` is now called `VCSEntry` (Probably should be called `VCSSource`)
  - The implementation of `Entry` / `VCSEntry` (with the yaml tags) for v1.0.0 was in the `lib/common` package. Not where things should be for implementation. Hence, it has moved to `lib/opencontrol/versions/1.0.0/opencontrol.go`
- `EntryDownloader` was renamed to `Downloader`
  - This was due to two facts:
    1. Too long of a name
    2. `Entry` no longer exists due to the last point.
- Clean up functionality between the `lib/opencontrol` and `commands/get`.
  - Prior to this patch, functionality used by the command/get was in the `lib` package. (`GetResources`, `GetRemoteResources`, `GetLocalResources`). **This made it very hard to separate out `lib/common`**
  - This logic has been moved to the `commands/get/resources` package
- Removed the `lib/opencontrol/versions/base` package
- Moved the interface `OpenControl` from `lib/opencontrol/versions/base/` to `lib/common`
  - Also, changed the interface methods from `Parse`, `GetSchemaVersion` and `GetResources` to `GetCertifications` `GetStandards` `GetComponents` `GetCertificationsDependencies` `GetStandardsDependencies` `GetComponentsDependencies`
  - This conforms with the other three yaml interfaces in which the interface returns information about the schema and not actions the schema should be doing. This makes it easier to separate between `lib` and `command` (which is also nice for plugins) 
- Removed `Worker` struct. It created a circular dependency when trying to use with mocks.


lib/opencontrol/versions/1.0.0/opencontrol_test.go
```diff
@@ -73,119 +31,52 @@ dependencies:
...
-			certifications, standards, components []string
-			worker *base.Worker
-			dependencies Dependencies
-			destination = "."
-			expectedError error
-			s Schema
+})
+
+
+var _ = Describe("Entry", func() {
```
jcscottiii: `VCSEntry`

commands/get/resources/downloader_test.go
```diff
@@ -7,27 +7,29 @@ import (
...
 			downloader := NewVCSDownloader()
 			assert.IsType(GinkgoT(), vcsEntryDownloader{}, downloader)
 		})
 	})
 	Describe("Downloading Entry from VCS", func(){
-		table.DescribeTable("DownloadEntry", func(e common.Entry, err error) {
+		table.DescribeTable("DownloadRepo", func(err error) {
+			remoteSource := new(commonMocks.RemoteSource)
+			remoteSource.On("GetURL").Return("")
+			remoteSource.On("GetRevision").Return("")
```
afeld: Why would these be empty?

commands/get/resources/getter.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	// Get Components
+	log.Println("Retrieving dependent components")
+	err = getter.GetRemoteResources(destination, constants.DefaultComponentsFolder, opencontrol.GetComponentsDependencies())
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
```
afeld: Can we split this up?

commands/get/resources/getter.go
```diff
@@ -0,0 +1,145 @@
...
+
+
+// Getter is an interface for how to get and place local and remote resources.
+type Getter interface {
+	GetLocalResources(source string, resources []string, destination string, subfolder string, recursively bool, resourceType constants.ResourceType) error
```
afeld: Maybe an options struct would make sense here?

commands/get/resources/getter.go
```diff
@@ -0,0 +1,145 @@
...
+	GetRemoteResources(destination string, subfolder string, entries []common.RemoteSource) error
+}
+
+// NewVCSAndLocalGetter constructs a new resource getter with the type of parser to use for the files.
+func NewVCSAndLocalGetter(parser opencontrol.SchemaParser) Getter {
+	return &vcsAndLocalFSGetter{Downloader: NewVCSDownloader(), FSUtil: fs.OSUtil{}, Parser: parser, ResourceMap: mapset.Init()}
```
afeld: Minor: Mind making this multiple lines? Will be a bit easier to visually scan everything that's being passed in.

commands/get/resources/getter.go
```diff
@@ -0,0 +1,145 @@
...
+			log.Printf("Copying local resource %s into %s\n", resourceSource, resourceDestination)
+			err = g.FSUtil.Copy(resourceSource, resourceDestination)
+		}
+		if err != nil {
+			log.Printf("Copying local resources %s failed\n", resourceSource)
+			return err
+		}
+	}
+	return nil
+}
```
afeld: Can we split this up?

commands/get/resources/getter.go
```diff
@@ -0,0 +1,145 @@
...
+		if err != nil {
+			return err
+		}
+		err = GetResources(tempPath, destination, opencontrol, g)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
afeld: Can we split this up?

commands/get/resources/getter_test.go
```diff
@@ -0,0 +1,155 @@
...
+			getter.On("GetLocalResources", "", components, destination, constants.DefaultComponentsFolder, true, constants.Components).Return(nil)
+			getter.On("GetRemoteResources", destination, constants.DefaultCertificationsFolder, dependentCertifications).Return(nil)
+			getter.On("GetRemoteResources", destination, constants.DefaultStandardsFolder, dependentStandards).Return(nil)
+			getter.On("GetRemoteResources", destination, constants.DefaultComponentsFolder, dependentStandards).Return(nil)
+		})
+		AfterEach(func() {
+			err := GetResources("", destination, s, getter)
+			assert.Equal(GinkgoT(), expectedError, err)
+			getter.AssertExpectations(GinkgoT())
+		})
```
afeld: Would `DescribeTable` be a more idiomatic way of doing this?

commands/get/resources/getter_test.go
```diff
@@ -0,0 +1,155 @@
...
+			getter.Downloader = downloader
+
+			schema := new(mocks.OpenControl)
+
+			parser := new(parserMocks.SchemaParser)
+			parser.On("Parse", mock.Anything).Return(schema, parserError)
+			getter.Parser=parser
+
+			err := getter.GetRemoteResources("dest", "subfolder", entries)
+			assert.Equal(GinkgoT(), expectedError, err)
```
afeld: Would help to break some of this out to helper functions.

lib/opencontrol/parse.go
```diff
@@ -19,26 +20,30 @@ const (
 	ErrMalformedBaseYamlPrefix = "Unable to parse yaml data"
 )
 
+// YAMLParser is the concrete implementation of parsing different schema versions in YAML format.
+type YAMLParser struct{}
+
 // Parse will try to parse the data and determine which specific version of schema to further parse.
-func Parse(parser base.SchemaParser, data []byte) (base.OpenControl, error) {
+func (parser YAMLParser) Parse(data []byte) (common.OpenControl, error) {
```
afeld: Should this be `func (parser *YAMLParser)`? Pardon my Go newb-ness.

lib/opencontrol/versions/1.0.0/opencontrol_test.go
afeld: `/me is assuming that's a TODO for yourself...?`

afeld:
Phew, ok, a handful of notes. Let's please please please try and keep the pull requests smaller for reviewer sanity 😁

commands/get/resources/downloader_test.go
jcscottiii: these don't matter as much since it's not actually downloading something. but i can put something here so people know what to expect!

commands/get/resources/getter.go
jcscottiii: will do!

commands/get/resources/getter.go
jcscottiii: will do!

commands/get/resources/getter_test.go
jcscottiii: unfortunately no due to this assertion `getter.AssertExpectations(GinkgoT())`

if i want to set the return values for the 3 calls to `GetLocalResources` and the 3 calls to `GetRemoteResources`, i need to make sure all 6 calls will be hit currently due to that assertion which checks to make sure your mocks were actually called. 

for example, the first case, i return an error on the very first call.
```go
		It("should return an error when it's unable to get local certifications", func() {
			expectedError = errors.New("Cert error")
			getter.On("GetLocalResources", "", certifications, destination, constants.DefaultCertificationsFolder, false, constants.Certifications).Return(expectedError)
		})
```
i can't add the other 5 calls because since the first one returns an error, the other functions aren't called and when that assertion is run, it will fail.

afeld: Hmm. Not sure that I _totally_ follow, but seems like it might be worth an inline comment?

jcscottiii: will do!

lib/opencontrol/versions/1.0.0/opencontrol_test.go
jcscottiii: ^yep!

commands/get/resources/getter.go
jcscottiii: will do!

commands/get/resources/getter.go
jcscottiii: will do! (this one is on the same line as on https://github.com/opencontrol/compliance-masonry/pull/217#r75931392)

commands/get/resources/getter_test.go
jcscottiii: will do!

commands/get/resources/getter.go
jcscottiii: done

commands/get/resources/getter.go
jcscottiii: done

lib/opencontrol/parse.go
jcscottiii: not here because there are no member variables of YAMLParser that are being modified that we want to preserve. Let's look at the declaration of the struct
```go
type YAMLParser struct{}
```

Now, if we want to, for example, keep track of the number times we parse with the parser, our declaration would look like 

```go
type YAMLParser struct{
    Count int
}
```

Now, if we didn't have the pointer (as-is now)
```go
func (parser YAMLParser) Parse {
    // stuff
    parser.Count++
}
```
Count would remain 0 forever.

and if you ran with the pointer, the count would update.
Example: https://play.golang.org/p/F-OVb1uVE2


tl;dr There's no state that needs to be carried between Parsing

commands/get/resources/getter_test.go
jcscottiii: actually, i can make this into a table testing if i remove the assertion. that assertion is helpful for making sure order of execution is important. but if we eventually start parallelizing some of these, we could get rid of that because order wouldn't matter.

jcscottiii: tl;dr will move to using DescribeTable!

commands/get/resources/getter.go
jcscottiii: if there were multiple fields (more than 1) that were used commonly between the two interface methods, i would do it.

jcscottiii:
@afeld this should be ready!

commands/get/resources/getter_test.go
```diff
@@ -0,0 +1,186 @@
...
+	"github.com/opencontrol/compliance-masonry/lib/common/mocks"
+	"github.com/opencontrol/compliance-masonry/lib/opencontrol"
+	parserMocks "github.com/opencontrol/compliance-masonry/lib/opencontrol/mocks"
+	"github.com/opencontrol/compliance-masonry/tools/constants"
+	"github.com/opencontrol/compliance-masonry/tools/fs"
+	fsmocks "github.com/opencontrol/compliance-masonry/tools/fs/mocks"
+	"github.com/opencontrol/compliance-masonry/tools/mapset"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+)
```
afeld: This many imports seems like a red flag that the setup for these structs/functions are too complicated, though I don't have specific suggestions about how to simplify 😕

PR #225, opened 2016-09-29
Go Vet, Go Fmt, and Golint

jcscottiii:
A cleanup patch. Since codeclimate won't run on PRs that target non-master targets, figured we need to fix that all now

Builds on #217


jcscottiii:
@afeld this is good. there needs to be another PR for configuring the `patch` parameter. this PR doesn't change any logic.

PR #229, opened 2016-10-06
shadowed `err` variable created

jackielii:
it makes the return value always nil


jcscottiii:
great catch! thanks for your contribution!

PR #238, opened 2016-10-20
Bump version to v1.1.3

jcscottiii:
This is cut a release that can read component v3.1.0


afeld:
We shouldn't release, since there _seems to_ be a regression: https://github.com/opencontrol/compliance-masonry/issues/234. Sorry!

jcscottiii:
Good catch! just wrote up a diagnosis to that bug

PR #251, opened 2017-03-31
ensure that errors are handled

afeld:
* Run [errcheck](https://github.com/kisielk/errcheck) through Circle CI
* Ensure that errors are handled in code, or at least explicitly ignore them (in tests)

### DIFF
--- a/lib/workspace.go
+++ b/lib/workspace.go
@@ -57,10 +57,14 @@ func LoadData(openControlDir string, certificationPath string) (common.Workspace
 		certificationErr = ws.LoadCertification(certificationPath)
 	}(&wg)
 	wg.Wait()
+
 	var errs []error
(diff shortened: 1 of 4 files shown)


lib/workspace.go
```diff
@@ -57,7 +58,7 @@ func LoadData(openControlDir string, certificationPath string) (common.Workspace
 	}(&wg)
 	wg.Wait()
 	var errs []error
-	// errs = append(errs, certificationErr)
+	errs = append(errs, certificationErr)
```
afeld: Hmm, this seems to be causing the build failure... @jcscottiii Happen to remember why it was commented out?

jcscottiii: ahh we need to do a check if certificationErr is not nil. if so, append it.

PR #254, opened 2017-04-26
Update Linux install commands with sudo

shawndwells:
On enterprise linux distros, the /usr/local/bin/ directory is protected from non-privileged users. Need sudo to copy files into that directory.

`````
$ cp compliance-masonry_1.1.2_linux_amd64/compliance-masonry /usr/local/bin/
cp: cannot create regular file ‘/usr/local/bin/compliance-masonry’: Permission denied
`````

This patch updates the install example to use sudo.


afeld:
Thanks!

PR #259, opened 2017-07-13
Update standard v1.0.0 to include "description" field

geezyx:
Related:
https://github.com/opencontrol/schemas/pull/56
https://github.com/opencontrol/schemas/issues/60

Based on a slack conversation with @anweiss, this PR adds compliance-masonry support for a "description" field for standards, based on the v1.0.0 schema.  [NIST-800-53 standards](https://github.com/opencontrol/NIST-800-53-Standards/blob/master/NIST-800-53.yaml) have already added the field as an example.

This PR also adds the description to the top of each standard page in the generated gitbook documentation.  For example:

image

### DIFF
--- a/commands/docs/gitbook/gitbookCertification.go
+++ b/commands/docs/gitbook/gitbookCertification.go
@@ -90,11 +90,15 @@ func (openControl *OpenControlGitBook) getControlOrigin(text string, controlOrig
 func (openControl *OpenControlGitBook) exportControl(control *ControlGitbook) (string, string) {
 	key := replaceParentheses(fmt.Sprintf("%s-%s", control.standardKey, control.controlKey))
 	text := fmt.Sprintf("#%s\n##%s\n", key, control.GetName())
+	if len(control.GetDescription()) > 0 {
+		text += "#### Description\n"
+		text += control.GetDescription()
+	}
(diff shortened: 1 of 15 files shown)


geezyx:
A couple additional notes:

* I updated fixtures/opencontrol_fixtures/standards/NIST-800-53.yaml with the latest from https://github.com/opencontrol/NIST-800-53-Standards/blob/master/NIST-800-53.yaml
* Added an extra line break to ensure proper markdown formatting -
 https://github.com/opencontrol/compliance-masonry/pull/259/files#diff-20fcfd66fc0414154e35c8266e13e17aR101
* Updated tests to expect the new output based on the two changes above

anweiss:
LGTM ... @jcscottiii ^

jcscottiii:
Thanks so much for this @geezyx !!

geezyx:
@jcscottiii np!  Question - what would constitute a schema version bump?  Should the "standard" become v1.0.1 with this change, or something similar?

mogul:
So great to see this... Thanks!

jcscottiii:
@geezyx this one doesn't need a version bump. Our current standard schema had the `description` field but compliance masonry didn't have it. 
One thing that needs to happen with the standard schema is actually add a version field. (But then CM would need to recognize when there's not a version field)

PR #260, opened 2017-07-15
switch to using dep for dependency management

afeld:
Also, removed the dependencies from the repository, to force us to be explicit in the manifest. When reviewing the diff, you can ignore once it starts to get into `vendor/`, as all of those files are deleted, and there are none modified that are lower in the alphabet.


jcscottiii:
I created an issue regarding moving off glide in appyveyor (we need it for glide novendor)

Issue #257, opened 2017-04-27
Fix reursively mispelling


jcscottiii:
There's a misspelling. Reurisvely -> Recursively.
https://github.com/opencontrol/compliance-masonry/blob/6513ee265cf7886e7ead1096d478e9fd93dd0a41/commands/get/resources/getter.go#L134

A great way for someone to dive into the code.

PR #265, opened 2017-07-26
Fix reursively spelling.

phillbaker:
Fixes https://github.com/opencontrol/compliance-masonry/issues/257.

(Doesn't occur elsewhere in the codebase.)


afeld:
Thanks!

PR #267, opened 2017-08-18
Support 'export' option

andybrucenet:
Goal: Output from `compliance-masonry` can be pipelined into a toolchain such as `docxtemplater`.

Process:

1. Create local `compliance` folder containing `opencontrols.yaml`, local policies, possibly local copy of SSP Template to use.
2. Run ‘compliance-masonry get’ as per usual
3. Run ‘compliance-masonry export’ – the new verb I added. Output can be to STDOUT or a result document in either JSON or YAML, flattened or not.

"Flatten" Notes:

The `export` command supports `--flattened` - so what does that mean? Basically, consider that a project could include components like RHEL7, OpenShift-v3, as well as project-specific (e.g. AU_policy component). In the generated output, the satisfied controls narratives / properties are located under the specific component (e.g. AC-2 (9) perhaps under RHEL7, AU-3 maybe under OpenShift-v3, and possibly AU-1 under local AU_policy component). This makes setting up a generic SSP Word template impossible, because there is no way to know a priori what flattened key a specific satisfied control would have. To handle this, the `export` verb looks for all satisfied controls and builds some meta-control definitions like `controls:AU-1:XXX` where the `XXX` is simply all the attributes stolen from the appropriate compliance-masonry output.

Hope this is a useful idea / PR!


andybrucenet:
Removed the extraneous output from export tests.

shawndwells:
@andybrucenet not sure if you're still interested, but would be great to see this PR rebased. Lots of great capability in this!

shawndwells:
Hi @andybrucenet - Much of this was incorporated into https://github.com/opencontrol/compliance-masonry/pull/292. What do you think? Should export be extended further?

Going to close this PR for now. New ones that extend ``export`` would be great!

Issue #270, opened 2017-10-25
go-utils/ufs dependency not found


gbrunett:
Please forgive the newbie question., but is anyone else getting this error?  Is there a workaround or fix available?

$ go get github.com/opencontrol/compliance-masonry
# cd .; git clone https://github.com/go-utils/ufs /usr/local/bin/src/github.com/go-utils/ufs
Cloning into '/usr/local/bin/src/github.com/go-utils/ufs'...
fatal: could not read Username for 'https://github.com': terminal prompts disabled
package github.com/go-utils/ufs: exit status 128

Any guidance would be greatly appreciated.

jamesrenfro:
I'm seeing the same error - it seems that ufs might be a private repo.

jcscottiii:
it used to be a public repo. seems like they removed it / made it private. need to move to another library that abstracts file system operations. maybe something like: https://github.com/spf13/afero

afeld:
Fixed by #272. Sorry for the delay!

PR #272, opened 2018-01-28
Replaced 3 calls to go-utils/ufs with metaleap/go-util 

tobybot11:
Replaced the 3 calls to go-utils/ufs with metalapp/go-util.
Only difference was adding an empty string in call to CopyAll for the 4th skipFileSuffix parameter.
Renamed the import to meatleapfs to prevent overlaps.

Complied and Tested on macOS High Sierra 10.13.2

∴ go version
go version go1.9.3 darwin/amd64

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -33,37 +33,59 @@
 
 [[projects]]
   branch = "master"
-  name = "github.com/go-utils/ufs"
-  packages = ["."]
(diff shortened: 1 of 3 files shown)


afeld:
Fixes #270.

afeld:
Thanks, and sorry for the delay!

PR #277, opened 2018-05-30
Add RPM spec file

redhatrises:


shawndwells:
@redhatrises can you add docs on using the spec to build RPM?

shawndwells:
OK. Created a ticket to track that action.

PR #279, opened 2018-05-30
Update versioning and add version const

redhatrises:


shawndwells:
ACK'ing, with understanding the .goxc.json will (eventually) be going away.

PR #281, opened 2018-05-31
Update tests

### LINKED ISSUE #262: Upgrade to Go 1.9 in CI/CD systems
~currently, we need glide to run `glide novendor` for tests. On *nix systems, we don't need to do that.~

~Check out the appyveyor.yml~

redhatrises:
- Remove dotdotdot in go tests for go-1.9 compatibility
- Update circleci tests to go-1.10 as AppVeyor tests already do this
- Update the circleci config to match latest configurations
- Remove Go dependencies as test should fail if dependency is missing from `vendor/` folder
- Fixes #262

### DIFF
--- a/.circleci/README.md
+++ b/.circleci/README.md
@@ -1,4 +1,4 @@
 ## Circle CI Scripts
 
 Call these scripts from the root of the repository
-- `circleci/coverage.sh # example`
+- `./.circleci/coverage.sh # example`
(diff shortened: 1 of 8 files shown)


redhatrises:
circleci tests fail because there is no error handling in the code for `App.Writer` and `c.App.Writer` which is a valid error.

shawndwells:
Thanks, @redhatrises !

PR #282, opened 2018-05-31
Remove Go Deps

redhatrises:
- vndr tool should be used instead as it is much simpler


shawndwells:
Thanks for rebasing. Ack!

PR #288, opened 2018-06-05
Use GORELEASER

redhatrises:
- GoReleaser is a release automation tool for Go projects, the goal is to simplify the build,
  release and publish steps while providing variant customization options for all steps.
  See https://goreleaser.com/

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,32 @@
+builds:
+  - binary: compliance-masonry
+    main: ./masonry-go.go ./diff.go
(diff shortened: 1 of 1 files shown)


shawndwells:
Have not used goreleaser before and was reading through it's quick start.

https://goreleaser.com/quick-start/

Should we add an ``archive`` target? e.g.
`````
archive:
  format: tar.gz
  replacements:
    amd64: 64-bit
    darwin: macOS
    linux: Tux
  files:
    - LICENSE.md
`````

shawndwells:
The ``archive`` target can be added later. Not blocking this PR because of it.

PR #291, opened 2018-06-08
Rename executable to masonry

### LINKED ISSUE #177: proposal: rename executable to `masonry`
I don't think that name is taken by any major tool, and it's faster to type than `compliance-masonry`. If, however, it makes lots of tools that assume an executable of the package name harder to use, it's not worth it.

redhatrises:
- Create a new compliance-masonry executable for legacy uses
- New masonry.go is now the primary program naming
- Fixes #177

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,8 +11,11 @@ jobs:
       - checkout
 
       - run:
-          name: Build the program
+          name: Build the compliance-masonry program
           command: go build cmd/compliance-masonry/compliance-masonry.go
+      - run:
+          name: Build the masonry program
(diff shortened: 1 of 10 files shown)


shawndwells:
Manually reviewed codecov failure. Ignoring. While testing coverage didn't increase, this is part of general refactoring.

shawndwells:
Good call on keeping the older compliance-masonry around during a transition time. Thanks for cleaning the duplicate errors in main()

Because this has been talked about several times -- in the referenced issue, at conferences, general conversations, going to merge this rename. Can start a discussion on if the entire repo should be renamed to match.

PR #292, opened 2018-06-11
Add export option

redhatrises:
- Fixes #267

### DIFF
--- a/pkg/cli/export/export_config.go
+++ b/pkg/cli/export/export_config.go
@@ -1,0 +1,114 @@
+package export
+
+import (
+	"bytes"
+	"encoding/json"
+
+	lib_certifications "github.com/opencontrol/compliance-masonry/pkg/lib/certifications"
+	"github.com/opencontrol/compliance-masonry/pkg/lib/common"
+)
+
+////////////////////////////////////////////////////////////////////////
+// Package structures
+
+// Config contains settings for this object
+type Config struct {
+	// remainder are configuration settings local to Export
+	Certification   string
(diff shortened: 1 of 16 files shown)


shawndwells:
@redhatrises https://github.com/opencontrol/compliance-masonry/pull/291 forced a rebase of this

pkg/cli/export/export_config.go
```diff
@@ -0,0 +1,114 @@
...
+
+	// certification
+	buffer.WriteString("\"certification\":")
+	bytesJSON, err := lib_certifications.MarshalJSON(p.Certification)
+	if err != nil {
+		return nil, err
+	}
+	buffer.WriteString(string(bytesJSON))
+
+	// iterate over components
```
shawndwells: ``components`` and ``standard`` iterations are very similar. Could this be condensed into a for each loop?

e.g.
`````
    arrayOpenControl := [2]string{"components", "standards"}

    for index,element := range arrayOpenControl{
        if len(p.element) > 0 {
           .... do stuff
        }     
    }   
}
`````

redhatrises: @shawndwells this actually would probably be a separate function as it deals with more than just strings.
I was hoping to avoid code cleanup in these type of PRs because of how big they are.

redhatrises: Tracked in #295

shawndwells: Sounds good. Thanks for opening the ticket to track.

PR #303, opened 2018-06-18
Make sure to account if compliance-masonry path is in a subdirectory

redhatrises:
Fixes the issue when running the `compliance-masonry` command like so:

- `./somedir/compliance-masonry`
- `../../../somedir/compliance-masonry`


shawndwells:
very helpful for devs who have messed up gopaths *ahem*

ack

Issue #170, opened 2016-06-14
As a developer, I should be able to use compile-time flags to create a version


jcscottiii:
Instead of creating a commit for just bumping the version in the code, you can use Go's `ldflags` to input a version at compile time. (You already have to bump the version in .goxc.json)

Search for it here: https://github.com/laher/goxc

http://www.atatus.com/blog/golang-auto-build-versioning/

Not sure but I think goxc can take PackageVersion and add it to ldflags automatically.

shawndwells:
resolved through https://github.com/opencontrol/compliance-masonry/pull/304

PR #304, opened 2018-06-18
goreleaser updates

redhatrises:
- Update goreleaser
- Add usage template
- Allow Makefile and goreleaser to build version into releases


redhatrises:
This should also fix #170

.goreleaser.yml
```diff
@@ -22,10 +33,10 @@ nfpm:
...
 brew:
-  name: oscalkit
+  name: compliance-masonry
   github:
     owner: opencontrol
-    name: homebrew-oscalkit
+    name: homebrew-compliance-masonry
   commit_author:
     name: OpenControl
     email: dev@open-control.org
```
shawndwells: Does this EMail exist?

cmd/masonry/masonry_test.go
```diff
@@ -12,19 +12,24 @@ import (
...
+Commands:
   diff        Compliance Diff Gap Analysis
   docs        Create compliance documentation
   export      Export to consolidated output
   get         Install compliance dependencies
   help        Help about any command
+  version     Display version
 
-Flags:
+Options:
```
shawndwells: given the ``usage`` output, should this be renamed to ``command-options``?

cmd/masonry/masonry_test.go
```diff
@@ -12,19 +12,24 @@ import (
 
 var usage = `
 Usage:
-  masonry [command]
+  masonry [global-options] COMMAND [command-options]
```
shawndwells: where are the ``global-options`` documented?

redhatrises: @shawndwells In the help/usage outputs for each of the commands. For example, `masonry help diff`

cmd/masonry/masonry_test.go
redhatrises: No. It shouldn't.

.goreleaser.yml
redhatrises: @shawndwells figured you would know more than I about that.

cmd/masonry/masonry_test.go
shawndwells: Bah

shawndwells: I mean you're right.

default ``help`` behavior vs ``help get``

.goreleaser.yml
shawndwells: Not something that's being changed in the PR, so no reason to hold a merge, just caught my eye.

PR #305, opened 2018-06-19
Fix Export typos

redhatrises:
- typos were causing crashes

### DIFF
--- a/pkg/cli/export/export.go
+++ b/pkg/cli/export/export.go
@@ -49,12 +49,12 @@ func RunExport(out io.Writer, cmd *cobra.Command, args []string) error {
 	}
 
 	// read parms
-	parmOpencontrols := cmd.Flag("opencontrols").Value.String()
-	parmDestination := cmd.Flag("destination").Value.String()
+	parmOpencontrols := cmd.Flag("opencontrol").Value.String()
+	parmDestination := cmd.Flag("dest").Value.String()
 	parmOutputFormat := cmd.Flag("format").Value.String()
-	parmFlatten := false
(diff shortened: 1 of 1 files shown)


pkg/cli/export/export.go
```diff
@@ -49,12 +49,12 @@ func RunExport(out io.Writer, cmd *cobra.Command, args []string) error {
 	}
 
 	// read parms
-	parmOpencontrols := cmd.Flag("opencontrols").Value.String()
-	parmDestination := cmd.Flag("destination").Value.String()
+	parmOpencontrols := cmd.Flag("opencontrol").Value.String()
```
shawndwells: should this be renamed to ``parmOpencontrol`` to match?

redhatrises: @shawndwells doesn't have to be, but can if you want it.

shawndwells: The naming should be aligned.

Was looking at how it's used:
`````
	// construct args
	config := Config{
		Certification:   args[0],
		OpencontrolDir:  parmOpencontrols,
		DestinationFile: parmDestination,
		OutputFormat:    outputFormat,
		Flatten:         parmFlatten,
		InferKeys:       parmInferKeys,
		Docxtemplater:   parmDocxtemplater,
		KeySeparator:    parmKeySeparator,
	}
`````

Should it really be called something like ``parmInputDir``?

Minimally should be ``parmOpencontrol`` (singular) to match the command flag. ``parmInputDir`` would be more descriptive.

What do you think?

Not the most important thing in the world, but while this part of the code is being touched....

PR #308, opened 2018-06-20
Enable automating binary builds with goreleaser when new tag is pushed

redhatrises:


shawndwells:
Thanks - getting closer to being able to do the auto builds!

PR #317, opened 2018-07-13
Fix GIT version in Makefile

redhatrises:
Now that `git tag` is done properly, we should get the version with `git describe`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ BUILD_DIR ?= _output
 BUILD_INFO := $(shell date -u +"%Y-%m-%dT%T%z")
 BUILD_PLATFORMS = "windows/amd64" "windows/386" "darwin/amd64" "darwin/386" "linux/386" "linux/amd64" "linux/arm" "linux/arm64"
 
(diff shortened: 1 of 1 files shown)


shawndwells:
@redhatrises the old and new way report different versions. Is this intended?

`````
$ git for-each-ref --format="%(refname:short)" --sort=-authordate --count=1 refs/tags | cut -d"v" -f2
1.1.3

$ git describe --abbrev=0 --tags | cut -d"v" -f2
1.1.4
`````

redhatrises:
@shawndwells yes... the latest version is `1.1.4` not `1.1.3`

shawndwells:
Wasn't the question. The old way returned $ver-1, new way returns $ver.

Does this mean the old method always had a bug that is just now being exposed?

redhatrises:
@shawndwells version 1.1.3 wasn't tagged correctly hence the old way was added. Now that version 1.1.4 was tagged correctly, this change uses the correct way of getting the tag for versioning.

shawndwells:
thanks @redhatrises !

PR #324, opened 2018-07-31
Remove metaleap code and replace with built in code

redhatrises:
- Metaleap plugins aren't 32bit which is causing build failures, and only 3 functions are really used from Metaleap. Repo is also abandoned. This removes the plugins and adds "like" functionality.


shawndwells:
thanks @redhatrises !

PR #325, opened 2018-07-31
Use GO stdlib errors instead of vektra errors

redhatrises:


shawndwells:
Good to see standard errors being used vs 3rd party plugins! Thanks for this!

Looks good once merge conflicts taken care of.

PR #328, opened 2018-08-16
update AppVeyor build key

shawndwells:
creates key with appropriate repo permissions


shawndwells:
self ack to get the v1.1.5 release out

PR #334, opened 2018-11-15
Added rational sort order for Justifications, Verifications, and Components

riskpeep:
Added natural sort for justifications and verifications.
Added sort by key for components

Provides rational sort order for generated gitbook documents.

PR #335, opened 2018-11-15
Fix sed package in appveyor

redhatrises:


shawndwells:
fixes issue from https://github.com/opencontrol/compliance-masonry/pull/334

PR #341, opened 2019-03-11
update docs to copy masonry binaries

shawndwells:


redhatrises:
@shawndwells can you remove `compliance-masonry` and just use `masonry`?

shawndwells:
On 3/11/19 11:11 AM, Gabe Alford wrote:
> @shawndwells <https://github.com/shawndwells> can you remove 
> |compliance-masonry| and just use |masonry|?


Yes, not a problem.

If the compliance-masonry binaries are no longer needed, should the 
build system be updated to remove (or not create) them? What's the 
reason both are included in the distribution? IIRC this had something to 
do with symlinks or backwards compatibility with people who still use 
compliance-masonry vs masonry.

redhatrises:
> On 3/11/19 11:11 AM, Gabe Alford wrote: @shawndwells <https://github.com/shawndwells> can you remove |compliance-masonry| and just use |masonry|?
> Yes, not a problem. If the compliance-masonry binaries are no longer needed, should the build system be updated to remove (or not create) them? What's the reason both are included in the distribution? IIRC this had something to do with symlinks or backwards compatibility with people who still use compliance-masonry vs masonry.

Yes, that's true, but maybe now is the time to remove it.

redhatrises:
I'm going to go ahead and ack. As really, there should be an update to the install section to not require updates every version.

shawndwells:
On 3/11/19 11:26 AM, Gabe Alford wrote:
> I'm going to go ahead and ack. As really, there should be an update to 
> the install section to not require updates every version.


Thanks!

Agree with the ack for a slightly different reason. If current releases 
have both binaries they should both be documented.

Future releases can drop the legacy names.

PR #344, opened 2019-03-27
Add search for implementation_status

timothy-spencer:
You now can search for different implementation_statuses like so:

`compliance-masonry info -i partial LATO`

Things were a bit old, so as a part of this, I migrated to dep instead of vndr, since that seems to be what the cool people are using, and I couldn't get vndr to just upgrade github.com/stretchr/testify.  This also brought a lot of other dependencies up to date.

I also fixed up a few tests that seemed to be broken, and added tests for the implementation status stuff.

Let me know what you think!


redhatrises:
Thanks for the PR @timothy-spencer! 

The only thing that I think makes sense to change is changing `implementationstatus` to `info`.
The reason is that I see this as a useful feature that is begging to be expanded to get all other sorts of information and stats like % complete, etc.

timothy-spencer:
Yeah, I have been thinking that it would be really cool to expand on this to allow for arbitrary queries, but that seemed hard.  How about I rename this to be info, and then think about how to add in query stuff (https://github.com/caibirdme/yql looks interesting) later on?

redhatrises:
> Yeah, I have been thinking that it would be really cool to expand on this to allow for arbitrary queries, but that seemed hard. How about I rename this to be info, and then think about how to add in query stuff (https://github.com/caibirdme/yql looks interesting) later on?

Sounds good to me. The thing to think about if you decide to do some querying stuff is how supported in the long run is the module itself e.g. YQL hasn't been updated in a year and not much active development.

timothy-spencer:
I'm not set on anything in particular.  That was just the best looking thing that popped up when I searched for "golang query language".  :-)  Let me know if you like this version better.  I'm happy to remove the signposts for how/where to implement generic queries if you like, but I thought that it might be useful later on.

pkg/cli/info/info.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/opencontrol/compliance-masonry/tools/constants"
+	"github.com/spf13/cobra"
+	"github.com/tg/gosortmap"
+)
+
+// NewCmdInfo allows you to query for implementation_status.
+func NewCmdInfo(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "info",
+		Short: "Compliance search",
```
redhatrises: @timothy-spencer let's make this `Get Compliance information from documentation` or something similar.

pkg/cli/info/info.go
```diff
@@ -0,0 +1,129 @@
...
+		Run: func(cmd *cobra.Command, args []string) {
+			err := RunInfo(out, cmd, args)
+			clierrors.CheckError(err)
+		},
+	}
+	cmd.Flags().StringP("opencontrol", "o", constants.DefaultOpenControlsFolder, "Set opencontrol directory")
+	cmd.Flags().StringP("implementation_status", "i", "", "implementation_status to search for")
```
redhatrises: `implementation_status` should be `implementation-status`

timothy-spencer:
I think that it would fulfill better the principle of least astonishment to have the flag match what it is searching for, but I'm fine either way.  Your changes have been plugged in.  Let me know what you think!

redhatrises:
Thanks. Ack

PR #345, opened 2019-04-23
gitbook pdf docs seem to be gone, replaced with ones that work

timothy-spencer:


afeld:
Looks like [PDF support was removed entirely](https://docs.gitbook.com/faq#does-gitbook-export-to-pdf). Maybe we take out that whole section?

timothy-spencer:
Well, it is interesting.  The version that you get from npm install still has it in it.  So I think it's still relevant for now.  I'm using it, anyways:  https://github.com/18F/gcp-appengine-template/blob/master/generate-compliance-pdf.sh

I suspect it's just using the old version.  Which is fine.  It is the #1 request for the gitbook folks (https://gitbook.canny.io/feature-requests), apparently, so it seems like a matter of time before they get that going again anyways.

timothy-spencer:
Just going through my outstanding PRs, and thought I'd see if this was stuck on your end, or if there was something else I needed to do.

PR #347, opened 2020-02-11
New Feature `masonry validate` command

isimluk:
Over the time, I have seen many malformed opencontrol contents. Here is very simple `masonry validate` command that will print out any problems identified in your opencontrol masonry.

How to test:

    $ cd MY_OPENCONTROL_DIR/
    $ masonry get
    $ masonry validate

### DIFF
--- a/pkg/cli/validate/validate.go
+++ b/pkg/cli/validate/validate.go
@@ -1,0 +1,20 @@
+package validate
+
+import (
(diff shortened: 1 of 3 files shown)


isimluk:
Codeclimate findings seems to be incorrect. Do not know how I can fix those. Please advise! Thanks!

PR #356, opened 2020-08-25
Update Golang build version to 1.15

redhatrises:


.travis.yml
```diff
@@ -6,7 +6,9 @@ env:
 matrix:
   include:
     - os: osx
-      osx_image: xcode8.3
+      osx_image: xcode9.4
+      language: go
+        - 1.15
```
aegershman, line 11: forgive my unsolicited review, but something looks a little bit funky with this key, it looks like `language` is both a map and a list here? I fumbled around on top of some travisci doc here:

- https://docs.travis-ci.com/user/languages/go/#specifying-a-go-version-to-use 
- https://travis-ci.community/t/how-to-use-stages-in-multi-language-builds/2084
- https://config.travis-ci.com/matrix_expansion (???)

and I think it might need to be broken out like this? maybe? Idk. Or maybe how it is is right, it just feels weird.

```suggestion
      language: go
      go:
        - 1.15
```

(apologies, hope this isn't a burden by doing unsolicited commentary)

redhatrises: @aegershman no worries. Thanks for the review! You are absolutely correct. I fixed the problem. Must have missed adding the go key as a mental slip.

Issue #358, opened 2020-08-26
consider leveraging dependabot


aegershman:
see: I believe this would be contingent on https://github.com/opencontrol/compliance-masonry/pull/355

In order to evaluate any outdated go dependencies & generally set up infrastructure to keep compliance-masonry up-to-date, consider leveraging [github-integrated dependabot](https://dependabot.com/go/). 

---

Posting a snippet of what the `dependabot.yml` file content **could** look like. Since it's so small I figure I'd put it within this same tracking issue to keep it simple (I didn't want to assume a PR would even be wanted in this case):

```yml
# .github/dependabot.yml
---
version: 2
updates:
  - package-ecosystem: gomod
    directory: "/"
    schedule:
      interval: weekly
    open-pull-requests-limit: 3
```

... I think, at least. I haven't tried it out yet. I could try it out on a temporary hard fork of this repo. I'll update this issue if I get to doing that sometime this week and seeing if it works as expected 👍 

Thanks for folks' time and consideration

redhatrises:
I think this is done. So closing.

Issue #357, opened 2020-08-26
general: consider leveraging github actions


aegershman:
In order to move tooling closer to this software repository and simplify automation tooling (debatable claim*), consider leveraging github actions for compliance-masonry.

specifically, this would be replacing travis-ci with github-actions for running codecov on PRs / for builds / potentially for release workflows, but "leveraging github actions" is a general consideration.

Thanks for folks' time 👍

redhatrises:
Actually was thinking about doing this after the bump to Golang 1.15.

PR #355, opened 2020-08-25
Move to using go modules

aegershman:
closes https://github.com/opencontrol/compliance-masonry/issues/350

This PR swaps the dependency management system from `dep` to `go mod`. The dependencies were automatically transferred over by `go mod` by it using the existing `Gopkg.toml` content.

It was done using the following commands from this blog post https://blog.golang.org/migrating-to-go-modules

```sh
go mod init github.com/opencontrol/compliance-masonry
go mod tidy
# remove Gopkg.toml and Gopkg.lock
# remove vendor/ directory
```

I'm not 100% confident whether the `vendor` directory should be removed, or if it should be re-vendored specifically using `go mod vendor`, so just giving that heads up

(loosely related sidenotes: this PR could be followed by an addition to .goreleaser (not in this PR, but somewhat related https://github.com/opencontrol/compliance-masonry/pull/354) to have `before` hooks of `go mod tidy`. It'd also be worth considering adding a dependabot listener to check for any `go mod` updates (can create different issue for that), and double-checking the entire build process is happy with using `go mod` for dependencies)

Thanks for anyone's time and consideration 👍 

EDIT should also update https://github.com/opencontrol/compliance-masonry/blob/master/docs/development.md

Issue #350, opened 2020-08-20
Consider leveraging gomodules from dep


aegershman:
In order to keep in alignment with [`go`'s dependency management approach](https://github.com/golang/dep/commit/87f309484f0da00d4559dbe0ae5698f589f9a0d0), `compliance-masonry` should consider [leveraging gomodules](https://blog.golang.org/migrating-to-go-modules) for the dependencies used (which is quite few (which is quite nice))

Not pressing, just logging for the sake of project tracking. Thanks for folks' time 👍

redhatrises:
Thanks for opening this.

You are right that the repo needs to be updated to support the latest gomodule instead of the legacy dep and other dependencies.
PRs are welcome.

PR #360, opened 2020-08-27
Dependabot for gomod updates

aegershman:
part of https://github.com/opencontrol/compliance-masonry/issues/358

contingent on:
- https://github.com/opencontrol/compliance-masonry/pull/355
- https://github.com/opencontrol/compliance-masonry/issues/350

see also:
- https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/
- similar issue noted in opencontainers https://github.com/opencontainers/runc/issues/2536

Introduces leveraging dependabot for analyzing gomod updates. However, this might be problematic with `vendor/` due to dependabot not currently supporting the ability to additionally perform `go mod vendor` as per this issue: https://github.com/dependabot/dependabot-core/issues/670 -- this appears to be being handled in other projects (https://github.com/vmware-tanzu/octant/blob/master/.github/workflows/dependabot-gomod.yaml) by having a github action workflow (see also: https://github.com/opencontrol/compliance-masonry/issues/357) which performs go mod vendor on the PR which dependabot opens, which might not be too awful. If nothing else, this would enable being notified of available updates, and gets the ball rolling to start analyzing changes

We can also set this lower to something like "max open PRs, 5" so dependabot doesn't explode (the goal being that it never would get that bad), let me know what you all think

Thanks for the time and consideration 👍

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+---
+version: 2
+updates:
+  - package-ecosystem: gomod
+    directory: "/"
+    schedule:
+      interval: daily
+    open-pull-requests-limit: 10


isimluk:
LGTM.

`make vendor` can be done as a separate github action. I just tried that in https://github.com/GoComply/metaschema/pulls?q=is%3Apr+is%3Aclosed

PR #371, opened 2020-09-24
Enable GH Actions for linux

redhatrises:
This allows testing to happen using GH actions. MacOS and Windows builds are failing their tests on different items, so they will have to be added later.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,24 +1,19 @@
-name: Go
+---
+name: build
 on: [push]
 jobs:
-  run:
+  build:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         os: [ubuntu-latest]
-    env:
-      OS: ${{ matrix.os }}
     steps:
       - name: Set up Go
         uses: actions/setup-go@v2
         id: go
 
       - name: Check out code into the Go module directory
-        uses: actions/checkout@v1
-
-      - name: Get dependencies
-        run: |
-          go get -v -t -d ./...
+        uses: actions/checkout@v2
 
       - name: Build compliance-masonry
         run: go build cmd/compliance-masonry/compliance-masonry.go
@@ -28,7 +23,7 @@ jobs:
 
       - name: Run tests
         run: |
-          go test $(go list ./...) -covermode=count -coverprofile=coverage.txt
+          go test -covermode=count -coverprofile=coverage.txt ./...
           go tool cover -func coverage.txt
 
       - name: Upload coverage to Codecov
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,38 @@
+name: Go
+on: [push]
+jobs:
+  run:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+    env:
+      OS: ${{ matrix.os }}
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        id: go
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v1
+
+      - name: Get dependencies
+        run: |
+          go get -v -t -d ./...
+
+      - name: Build compliance-masonry
+        run: go build cmd/compliance-masonry/compliance-masonry.go
+
+      - name: Build masonry
+        run: go build cmd/masonry/masonry.go
+
+      - name: Run tests
+        run: |
+          go test $(go list ./...) -covermode=count -coverprofile=coverage.txt
+          go tool cover -func coverage.txt
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1
+        if: matrix.os == 'ubuntu-latest'
+        with:
+          file: ./coverage.txt
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,11 +15,8 @@ jobs:
       - name: Check out code into the Go module directory
         uses: actions/checkout@v2
 
-      - name: Build compliance-masonry
-        run: go build cmd/compliance-masonry/compliance-masonry.go
-
-      - name: Build masonry
-        run: go build cmd/masonry/masonry.go
+      - name: Build compliance-masonry and masonry
+        run: go build -o _output/ ./cmd/...
 
       - name: Run tests
         run: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,10 +16,6 @@ jobs:
       - run:
           name: Run tests
           command: ./.circleci/coverage.sh
-      - run:
-          name: Send results to Codecov
-          shell: /bin/bash
-          command: bash <(curl -s https://codecov.io/bash)
       - deploy:
           name: goreleaser
           command: |


redhatrises:
@isimluk or @aegershman thoughts on this PR?

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
+name: Go
+on: [push]
+jobs:
+  run:
```
aegershman, line 4: stylistic, since the name of the overall workflow is related to "build"ing, consider the job name to be `build`, again, stylistics, not a big deal

```suggestion
  build:
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
...
+      OS: ${{ matrix.os }}
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        id: go
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v1
+
+      - name: Get dependencies
```
aegershman, line 19: since we're `vendor/`'ing dependencies in source control, do we need to do a `go get` at all?

from what I remember in https://github.com/opencontrol/compliance-masonry/pull/355 (or at least think I remember, who knows), in go 1.14+, `go` commands will assume `-mod=vendor` by default if the `vendor/` directory exists; so like in `go test`, `go build`, all that good stuff, it should use the `vendor/` dir by default

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
...
+jobs:
+  run:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+    env:
+      OS: ${{ matrix.os }}
+    steps:
+      - name: Set up Go
```
aegershman, line 12: also, very minor point, we should be able to optionally remove the `name` and `id` sections on some of the more 'boring' tasks, and just let it use defaults; the `name` and `id` of course is just stylistic, just putting it out there

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v1
+
+      - name: Get dependencies
+        run: |
+          go get -v -t -d ./...
+
+      - name: Build compliance-masonry
+        run: go build cmd/compliance-masonry/compliance-masonry.go
```
aegershman, line 24: this ought to work just fine; we might benefit from recycling the tooling paths for consistency-- by that I mean, we should be able to get away with doing `make build` here based on the Makefile (https://github.com/opencontrol/compliance-masonry/blob/master/Makefile), or by doing `goreleaser build` (assuming the `goreleaser` binary was made available to the host); it all does the same thing, but it's through the same 'build path' so it's more concise. just a suggestion in any case  👍  this should still work

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Build masonry
+        run: go build cmd/masonry/masonry.go
+
+      - name: Run tests
+        run: |
+          go test $(go list ./...) -covermode=count -coverprofile=coverage.txt
+          go tool cover -func coverage.txt
```
aegershman, line 32: I might be completely full of it, but I think we can get rid of doing the `go list ./...` command, and replace both of these with...

```sh
go test -coverprofile=coverage.txt -covermode=count ./...
```

I think, at least. That's based on what we experimented with in (https://github.com/opencontrol/compliance-masonry/pull/355#issuecomment-681483319) and linked to this `go` issue (https://github.com/golang/go/issues/19090). not a huge deal either way, just putting it out there for convenience

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Upload coverage to Codecov
```
aegershman, line 34: again, purely stylistic, feel free to ignore/throw it back, but [for consideration, we should be able to get away with removing the `name` key entirely](https://github.com/codecov/codecov-action#usage) and having github actions use the default identifier of the step-- but by the same token, the same codecov-action example page has it included, so it's [really just style, no worries either way](https://github.com/codecov/codecov-action#example-workflowyml-with-codecov-action)

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: codecov/codecov-action@v1
+        if: matrix.os == 'ubuntu-latest'
```
aegershman, line 36: syntax here looks good to me; I assume we'd keep this conditional to only perform codecov for `ubuntu-latest` even if the other OS's (windows,macos-latest) were included here, too? (It should be the same in any case regardless of OS I'd have to assume, no need to upload x3 times, right?). regardless, lgtm

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
+name: Go
+on: [push]
+jobs:
+  run:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+    env:
+      OS: ${{ matrix.os }}
```
aegershman, line 10: is the envvar export of `OS` used by `go` or `codecov`?

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
+name: Go
+on: [push]
+jobs:
+  run:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
```
aegershman, line 8: just for reference, including the github actions [hosted runner virutal-environments "installed softwares" list for ubuntu here, including `go` itself](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#go). just sharing the list for link reference.

this looks good to me 👍  future would presumably just be to plop in `macos-latest`, `windows-latest` in here as well. good stuff.

aegershman (commented):
commented on parts and pieces; sorry for blabbing all over it; won't be heartbroken in any way if comments aren't taken, just putting optional thoughts on it. in any case, I'm pro github-actions and reducing touchpoints of other systems, if possible 👍  lmk if there's anything else I can do to be helpful.

.github/workflows/build.yml
redhatrises: exactly

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
...
+    env:
+      OS: ${{ matrix.os }}
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        id: go
```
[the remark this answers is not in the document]
redhatrises: Do you think we should with the minor version? We don't have to update as it will always use the default.

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Check out code into the Go module directory
```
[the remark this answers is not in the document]
redhatrises: Personally like leaving the `name` key as it increases readability.

.github/workflows/build.yml
redhatrises: As previously mentioned, prefer to keep it for readability.

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: actions/checkout@v1
```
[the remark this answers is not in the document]
redhatrises: done

.github/workflows/build.yml
redhatrises: done

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
+name: Go
```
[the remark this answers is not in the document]
redhatrises: done

.github/workflows/build.yml
redhatrises: good point. Removed.

.github/workflows/build.yml
redhatrises: See comments around name already.

aegershman: yeah, apologies, I didn't need to make the same comment on all locations 🥴

.github/workflows/build.yml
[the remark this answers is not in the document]
aegershman: I support your decision either way; but yeah, after thinking a bit more on it, default of `latest` is good and practically as effective as using the minor version; it'd be an edge-case if a build fails as a result of 1.15 to 1.16, that in that case the problem should need to be addressed pretty quick, or just pop the `with: go-version: xyz` on there if it becomes a problem 🤷 until then, yes default is good 👍

aegershman (approved):
comments on some minor nitpicks `go test` / `go build` thoughts, but nothing _critical_. not sure about whether `go get` usage is necessary or desired since we're using `go vendor`-- but any of that could always be tweaked/fixed in later PRs if it doesn't work or changed later.

ultimately lgtm 👍

isimluk (approved):
:+1:

.github/workflows/build.yml
```diff
@@ -15,11 +15,8 @@ jobs:
       - name: Check out code into the Go module directory
         uses: actions/checkout@v2
 
-      - name: Build compliance-masonry
-        run: go build cmd/compliance-masonry/compliance-masonry.go
-
-      - name: Build masonry
-        run: go build cmd/masonry/masonry.go
+      - name: Build compliance-masonry and masonry
+        run: go build -o _output/ ./cmd/...
```
redhatrises, line 19: @aegershman @isimluk merged the multiple binary step to this command. Thoughts?

aegershman: I think this is fine for now 👍 I think it's good enough to be merged in, and this could always be revised on specifics. If I understand correctly, I don't think the current output of the built artifact is used, I get the impression it's just being used to determine if the cli can be built at all.

redhatrises: Yes, that's true. Okay. Thanks! Would you be willing to look at getting MacOS and Windows up and running?

PR #377, opened 2020-10-02
Enable GH Actions for macOS and Windows

redhatrises:


redhatrises:
@aegershman what do you think of this? Not getting failures elsewhere in manual. Thoughts?

aegershman:
@redhatrises are you using a Linux distribution when testing locally?

I'm normally a bit skeptical of a "works on my machine"-type situation, but this is unique because these failing tests are testing the I/O of the built binary as part of the test; so there are more "seams" where things could be a little funky with the underlying operating system. I can share the modified code later if desired to see if it works on ci/cd. But my initial thoughts are it has to do with differences in async scheduling. 

I tried this on my personal MacBook (MBP 2015, OS11.0 Beta (20A5384c)) and I'm getting the same errors as the github actions. My hypothesis is that it has to do with a race condition between when [`gexec.CleanupBuildArtifacts()` is executed here](https://github.com/opencontrol/compliance-masonry/blob/master/pkg/tests/test_masonry.go#L30) and the [`Eventually(...)` is executed here in the test suite](https://github.com/opencontrol/compliance-masonry/blob/master/cmd/masonry/masonry_test.go#L24) polling for a binary which no longer exists due to `cleanup` running. When I move the "cleanupbuildartifacts" execution _after_ the `eventually...should` in the test, I get consistent successful results. I also get results when I call `session.Wait()` immediately before `gexec.CleanupBuildArtifacts()` in the [`Masonry(...)` command call here](https://github.com/opencontrol/compliance-masonry/blob/f5e64fbb4caf81401ded5c0b39711dea37fae2af/pkg/tests/test_masonry.go#L28-L30)

these are just my initial thoughts, but I think we could either...

- not bother with running these i/o tests (at least on macOS, idk about Windows), and just assume that if the binary builds, we're okay; from what I understand, that's what's served this project for a while. I don't think `go test` has been ran on windows/mac systems, I think it's always been on ubuntu-based systems(?). `go build` works fine.
- have each test suite which performs some sort of i/o based test to have a "beforeSuite" which calls "masonry(...)", and an "afterSuite" which calls `gexec.cleanupbuildartifacs()`, and remove the `gexec.cleanupbuildartifacts()` line out of the "session builder" function of `Masonry()` in `pkg/tests/test_masonry`
- building on option 2, remove the `pkg/tests/test_masonry` function altogether and duplicate that `gexec.build(...)` logic into the respective test packages of other parts of the code which perform i/o tests... this is just idea 2, but expanded a bit more

that's just from a little experimenting on my macbook though, so I don't know if any of this is 100% accurate. I think the Windows test failures are a different error-- it looks like the "move" operation on the "coverage.txt" maybe didn't work as intended for the windows container

thoughts? hopefully this isn't too off the mark

redhatrises:
@aegershman if you want to submit fixes for the race condition, that would be great! Personally, I think that we should run tests on windows/macOS just in case they do discover something that does break with a binary build.

aegershman:
👍 I agree. I can submit something over the next couple days. I assume it completely negates any benefit from async tests with `Eventually()`, but you could try adding `session.Wait()` right before `gexec.CleanupBuildArtifacts()` in `test_masonry.go`; at least that's what gets this working on my MacBook. not saying that's what should be merged, but would be curious if it works as a fix for the macos env

redhatrises:
@aegershman @isimluk seemed to have fixed the tests issues. Please look and agree or disagree with the changes.

aegershman:
https://github.com/opencontrol/compliance-masonry/issues/378 so I created this issue as a follow-up in case there's a desire to merge this in and then come back to address it later. Apologies, I'm pretty busy at work at the moment so I take back what I said about putting out a PR "in the next few days" 😬  sorry about that. I can't commit to putting out a PR for redressing some of the tests to make them appropriately async-- but I wouldn't be opposed to merging this as-is, and then coming back to address this a bit later. I believe it's just a matter of setup/teardown execution order in test logic rather than anything deeply problematic in the core code itself. +1, however would also hold for a blessing from @isimluk too

isimluk:
Looks good to me! Thank You guys!

isimluk (approved):
:+1:

